PR #10, opened 2020-08-24
Simplied Recorder and Metric

opus111:
Simplified design of Recorder and Metric concepts.  Now Recorders are simple Callbacks that logs particular statistics.  Metric and Recorder have gone away.


src/recorder.jl
```diff
@@ -29,111 +21,99 @@ https://dev.fast.ai/learner#Recorder
...
-function log!(rec::Recorder, name::String, epoch::Int, batch::Int, value)
-#    println("$name,$epoch,$batch,$value")
-    rec.logs[name][epoch,batch] = value
+struct ValidateLossRecorder <: AbstractCallback
+    log::Array{Real,2}
 end
 
-Base.getindex(rec::Recorder,idx...) = rec.logs[idx[1]][idx[2],idx[3]]
-Base.setindex!(rec::Recorder,value,idx...) = rec.logs[idx[1]][idx[2],idx[3]] = value
+before_fit(lr::TrainLossRecorder,lrn::Learner, epoch_count, batch_size) = lr.log = zeros((epoch_count,batch_size))
```
darsnack, line 45: Looks like you have a copy-paste typo. Should be `lr::ValidateLossRecorder`.

opus111: Thanks.  Fixed

ToucheSir (commented):
I noticed all of the metrics here have a "Recorder" suffix. Would it help to add a new namespace  and/or abstract type such that these names can be trimmed?

opus111:
> I noticed all of the metrics here have a "Recorder" suffix. Would it help to add a new namespace and/or abstract type such that these names can be trimmed?

Agreed.  I like that.

There is also a lot of repeated code in recorder.jl. Is there a nice idiomatic way to improve that?

ToucheSir:
The duplication doesn't look too bad, but one way would be to extract the {Train,Validation} part into a field or type parameter. I also noticed that fast.ai uses a single class for their training and validation loss recorder&mdash;this could work as well (moreso for the loss than for the training phase agnostic metrics).

src/FastAI.jl
```diff
@@ -29,7 +28,11 @@ export valid
 
 export DummyCallback
 export ProgressCallback
-export Recorder
+
+export TrainLossRecorder
+export ValidateLossRecorder
+export TrainSmoothLossRecorder
+export ValidateSmoothLossRecorder
```
darsnack, line 35: I think @ToucheSir's comment about type names means that here we should export the module `Recorder` but not the types within it. Then in user code, a person needs to write `Recorder.TrainLoss` to get the training loss recorder. I agree that this will look cleaner than having "Recorder" at the end of each type name.

src/recorder.jl
```diff
@@ -7,133 +7,110 @@ Port of the FastAI Recorder to Julia
...
 
-Add `name` to `rec` to be tracked.
+Record a log of training loss
 """
-add!(rec::Recorder, name, epoch_count, batch_size) = rec.logs[name] = fill(nothing,epoch_count,batch_size)
+struct TrainLoss <: Recorder
+    log::Array{Real,2}     
+end
+TrainLoss()=TrainLoss([])
+batch_train_loss(lr::TraninLoss,lrn::AbstractLearner, epoch, batch, loss) = lr.log[epoch,batch] = loss
```
darsnack, line 41: Think this is a typo. Should be `TrainLoss` not `TraninLoss`.

src/recorder.jl
```diff
@@ -7,133 +7,110 @@ Port of the FastAI Recorder to Julia
...
+This design is significantly different from the original 
+Recorder and Metrics objects in Python
 
 https://github.com/fastai/fastai2/blob/master/fastai2/learner.py
 
 The documentation is copied from here
 
 https://dev.fast.ai/learner#Recorder
 =#
+module recorder
```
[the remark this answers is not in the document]
opus111: I really like
```
struct Recorder{S, F, T} <: AbstractCallback where S, F, T<:Real
    f::F
    log::Array{T, 2}
end

ValidateLoss() = Recorder{:Validate}(identity, Float32[])
batch_validate_loss(lr::Recorder{:Validate}, ...) = # do stuff
```
Can we use the same trick with Smoothing, like this
```
ValidateLoss() = Recorder{:Nothing,:Validate}(identity, Float32[])
SmoothValidateLoss() = Recorder{:Smooth,:Validate}(identity, Float32[])

batch_validate_loss(lr::Recorder{:Smooth,:Validate}, ...) = # do stuff with Smoother
```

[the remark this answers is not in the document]
darsnack: Yeah, but for smooth loss you don't need to specify `:Smooth`. The type parameter `F` in `Recorder` is the type of `f` which is the function applied to the `loss` before recording it. This type will be `Smoother` when you pass in `Smoother` as the function. So you can specialize on that:
```julia
struct Recorder{S, F, T} <: AbstractCallback where S, F, T<:Real
    f::F
    log::Array{T, 2}
end

TrainLoss() = Recorder{:Train}(identity, Float32[])
SmoothTrainLoss(alpha = 0.98) = Recorder{:Train}(Smoother(alpha), Float32[])
ValidateLoss() = Recorder{:Validate}(identity, Float32[])
SmoothValidateLoss(alpha = 0.98) = Recorder{:Validate}(Smoother(alpha), Float32[])

function _record_state!(lr::Recorder, epoch, batch, state)
    lr.log[epoch, batch] = lr.f(state)
end

# example specializing on :Train vs :Validate
batch_train_loss(lr::Recorder{:Train}, lrn::AbstractLearner, epoch, batch, loss) = _record_state!(lr, epoch, batch, loss)
batch_validate_loss(lr::Recorder{:Validate}, lrn::AbstractLearner, epoch, batch, loss) = _record_state!(lr, epoch, batch, loss)

# example specializing on Smoother vs identity
function before_fit(lr::Recorder, lrn::Learner, epoch_count, batch_size)
    lr.log = zeros(epoch_count, batch_size)
end
function before_fit(lr::Recorder{<:Any, Smoother}, lrn::Learner, epoch_count, batch_size)
    reset!(lr.f)
    lr.log = zeros(epoch_count, batch_size)
end
```

You can do various combinations of the `Recorder{S, F, T}` type parameters to specialize.

opus111:
Many thanks for all the suggestions here and in the "Common Code" thread.

I spent the week learning those techniques, and rewriting this code several ways.  In the end however, I went with my original design.  Type arguments did not save many lines of code, limited a Recorder to a log and preprocessing function, and were harder to understand for a non-Julian.  I thought the last was important if we want to attract Python developers and have them add their own Recorders.

However, I will definitely use type arguments in the future.  It is a really cool and elegant technique.

PR #21, opened 2021-03-28
new tutorial: Migration guide from fastai's python library to FastAI.jl

War-Eagl:
First of three part tutorial on migrating from Fastai's python library to FastAI.jl. This is supposed to beginner level and I've written down things that I felt crucial when I came from the Fastai's python's library.

### DIFF
--- a/docs/notebooks/Migration_Guide_beginner.ipynb
+++ b/docs/notebooks/Migration_Guide_beginner.ipynb
@@ -29,7 +29,7 @@
       "source": [
         "# Migration Guide from FastAI's python library to FastAI.jl - Beginner\n",
         "\n",
-        "If you are a user of Fastai's python library and want to migrate to FastAI.jl, this three part tutorial series will get you started. In this first part of tutorial we'll see about **Datasets**, **dataloaders** and **Learner** objects of FastAI.jl"
+        "If you are a user of Fastai's python library and want to migrate to FastAI.jl, this three part tutorial series will get you started. In this first part of tutorial, we'll look at **datasets**, **dataloaders** and **learner** in FastAI.jl"
       ]
     },
     {
--- a/docs/notebooks/Migration_Guide_beginner.ipynb
+++ b/docs/notebooks/Migration_Guide_beginner.ipynb
@@ -374,7 +374,7 @@
         "id": "bdx1u49NFxjj"
       },
       "source": [
-        "Learner is the object that binds the data loaders with the model and sets the other hyper parameters like loss function, optimizers, etc. There is `Learner()` function which is similar to fastai's `cnn_learner` function. The below python code can be translated as,\n",
+        "The learner is the object that binds the data loaders with the model and sets the other hyper parameters like loss function, optimizers, etc. There is `Learner()` function which is similar to fastai's `cnn_learner` function. The below python code can be translated as,\n",
         "\n",
         "```\n",
         "learn = cnn_learner(dls, resnet34, metrics=error_rate)\n",
--- a/docs/notebooks/Migration_Guide_beginner.ipynb
+++ b/docs/notebooks/Migration_Guide_beginner.ipynb
@@ -52,7 +52,7 @@
       "source": [
         "# Datasets\n",
         "\n",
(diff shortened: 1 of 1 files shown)


darsnack:
I feel like one important part of fast.ai is the [data block API](https://docs.fast.ai/tutorial.datablock). This tutorial would benefit from a mapping from that API to our API (MLDataPattern.jl + DataAugmentation.jl).

docs/notebooks/Migration_Guide_beginner.ipynb
```diff
@@ -0,0 +1,448 @@
...
+        "# DataLoaders"
+      ]
+    },
+    {
+      "cell_type": "markdown",
+      "metadata": {
+        "id": "fFmwN26ZA5wO"
+      },
+      "source": [
+        "Dataloaders are important aspects of fastai's python library. These objects are like assembly lines that feed materials to the machine. Likewise dataloaders feed data to the model. \n",
```
darsnack, line 193:
```suggestion
        "Dataloaders are important aspects of fastai's python library. Like an assembly line feeding materials to the next machine, dataloaders feed data to the model. \n",
```

docs/notebooks/Migration_Guide_beginner.ipynb
```diff
@@ -0,0 +1,448 @@
...
+    },
+    {
+      "cell_type": "markdown",
+      "metadata": {
+        "id": "EmDKNgh2A3Dp"
+      },
+      "source": [
+        "# DataLoaders"
+      ]
+    },
```
darsnack, line 186: My concern with this section is that it blurs the line between data containers and data iterators. For a migration guide, I would prefer to see an explanation of how in FastAI.jl, we have a single interface — `MLDataPattern.getobs` and `MLDataPattern.nobs` — for both data containers and data iterators. The only distinction is that data iterators like `eachbatch` or  `eachobs` also implement the [iteration interface](https://docs.julialang.org/en/v1/manual/interfaces/#man-interface-iteration) from Base.

This should be contrasted with fast.ai where there is a distinction between datasets and data loaders.

docs/notebooks/Migration_Guide_beginner.ipynb
```diff
@@ -0,0 +1,448 @@
...
+        "# Learner"
+      ]
+    },
+    {
+      "cell_type": "markdown",
+      "metadata": {
+        "id": "bdx1u49NFxjj"
+      },
+      "source": [
+        "Learner is the object that binds the data loaders with the model and sets the other hyper parameters like loss function, optimizers, etc. There is `Learner()` function which is similar to fastai's `cnn_learner` function. The below python code can be translated as,\n",
```
darsnack, line 377:
```suggestion
        "The learner is the object that binds the data loaders with the model and sets the other hyper parameters like loss function, optimizers, etc. There is `Learner()` function which is similar to fastai's `cnn_learner` function. The below python code can be translated as,\n",
```

docs/notebooks/Migration_Guide_beginner.ipynb
```diff
@@ -0,0 +1,448 @@
...
+      "cell_type": "markdown",
+      "metadata": {
+        "id": "fHifey18IVqf"
+      },
+      "source": [
+        "However, `Learn` method makes us to define every parameters manually. Like we should explicitly convert our data container object to `dataloader` object and pass them. \n",
+        "\n",
+        "Instead we can cut the complexity by using `methodlearner` function where everything can be passed as hyperparameters. The cool thing about `methodlearner` is it takes the data container object and implicitly converts it into dataloader, where the batch size, validation split can be customized if we need. Therefore the above code can be converted into, "
+      ]
+    },
```
darsnack, line 419: I would just delete this section. The comparison to `Learner` isn't exact, and I think `cnn_learner` is more closely compared to `methodlearner`.

darsnack (changes requested):
Left some comments. Overall, I think this is a good start, but I would prefer to see more of a focus on "translating" concepts instead of code snippets.

War-Eagl:
I corrected the typos and deleted those lines in Learner section. I also added some explanation in Dataloaders section. Does that make any sense  and do any justice to the tutorial?

Issue #23, opened 2021-04-02
Add `TableDataset`


darsnack:
We need a `TableDataset` data container to work with tabular data. This data container should implement the `getobs`/`nobs` interface from LearnBase.jl, and it should wrap any "table" that satisfies the interface from Tables.jl.

PR #26, opened 2021-04-05
added TableDataset

manikyabard:
Added TableDataset, getobs and nobs which works with Tables.jl interface implementations. Closes https://github.com/FluxML/FastAI.jl/issues/23


src/datasets/containers.jl
```diff
@@ -46,4 +48,36 @@ isimagefile(file::String) = occursin(IMAGEFILE_REGEX, lowercase(file))
 const IMAGEFILE_REGEX = r"\.(gif|jpe?g|tiff?|png|webp|bmp)$"
 
 
-## TODO: TableDataset
+struct TableDataset{T}
+    table::T #Should implement Tables.jl interface
+    TableDataset{T}(table::T) where T = Tables.istable(table) ? new{T}(table) : error("Object doesn't implement Tables.jl interface")
+end
+
+TableDataset(path::String) = TableDataset(DataFrame(CSV.File(path)))
```
darsnack, line 56: Reflecting Ari's comment from Zulip, maybe we can just use `CSV.File` here.

src/datasets/containers.jl
```diff
@@ -46,4 +48,36 @@ isimagefile(file::String) = occursin(IMAGEFILE_REGEX, lowercase(file))
...
+TableDataset(table::T) where {T} = TableDataset{T}(table)
+
+function LearnBase.getobs(dataset::FastAI.Datasets.TableDataset{T}, idx) where {T}
+    if Tables.rowaccess(dataset.table)
+        for (index, row) in enumerate(Tables.rows(dataset.table))
+            if index==idx
+                return row
+            end
+        end
```
darsnack, line 65: Iterating all the rows seems like a really unfortunate consequence of the Tables.jl interface. CSV.jl support "jumping ahead" to a certain row in the file. Does Tables.jl not have a similar interface?

src/datasets/containers.jl
```diff
@@ -46,4 +48,36 @@ isimagefile(file::String) = occursin(IMAGEFILE_REGEX, lowercase(file))
...
+    elseif Tables.columnaccess(dataset.table)
+        rowvals = []
+        for i in 1:(length ∘ Tables.columnnames)(dataset.table)
```
darsnack, line 68:
```suggestion
        for i in 1:length(Tables.columnnames(dataset.table))
```

src/datasets/containers.jl
```diff
@@ -46,4 +48,36 @@ isimagefile(file::String) = occursin(IMAGEFILE_REGEX, lowercase(file))
...
+            append!(rowvals, Tables.getcolumn(dataset.table, i)[idx])
```
darsnack, line 69: Should this be `push!` or `append!`?

darsnack (commented):
This looks like a good start to me. I left a couple comments. My main concern is that it seems like we have to iterate a bunch of rows to get the i-th row?

src/datasets/containers.jl
manikyabard: Yeah it should be `push!` only, considering single item is being added.

src/datasets/containers.jl
```diff
@@ -46,4 +48,36 @@ isimagefile(file::String) = occursin(IMAGEFILE_REGEX, lowercase(file))
...
+        end
+        return rowvals
+    else error("The Tables.jl implementation used should have either rowaccess or columnaccess.")
+    end
+end
+
+function LearnBase.nobs(dataset::FastAI.Datasets.TableDataset{T}) where {T}
+    if Tables.rowaccess(dataset.table)
+        return length(Tables.rows(dataset.table))
```
ToucheSir, line 78: I saw a couple of implementations (e.g. JDBC) where `length` is not defined on the result of `rows`. The question is whether or not we care about that. There's also the question of whether this might end up materializing more data than we want as a side effect of calling `rows` or `columnaccess`. Unfortunately AFAICT there's no API in Tables.jl for querying the size of a source (if it has one).

src/datasets/containers.jl
manikyabard: I don't think Tables.jl requires to support "jumping ahead" to a certain row, unless it is possible to do so for all objects implementing iteration interface.

If the table is row access, the only required method is `Tables.rows` which just needs to return an iterable of objects satisfying the `AbstractRow` interface. `AbstractRow` only requires methods for getting column values and names.

If the table is column access, the only required method is `Tables.columns` which needs to return an object implementing the `AbstractColumns` interface and having methods for getting a column object and column names.

src/datasets/containers.jl
manikyabard: Then is the only option to iterate through the rows for the size, incase the table doesn't support column access? If it is, then should I make the required changes?

src/datasets/containers.jl
manikyabard: Okay I can change it so that the default behaviour on getting a path is to use `CSV.File` object. It's just that I thought having a familiar API like DataFrames might be easier to work with, specially for people coming from Python fastai.

darsnack: That's true, but I think in the this case it is more important to be performant reading from disk. Most people won't be directly interacting with the underlying table.

darsnack:
Based on the discussion, I am wondering if we should have a `TableDataset` and a `CSVDataset`. Seems like just doing the former is leaving a lot of performance on the table. CSV.jl will allow us to avoid iterating rows for CSV files.

darsnack:
For data containers, it makes more sense to target these based on file type? For tables already loaded for disk, I'm now thinking it makes more sense to directly define `getobs` on the Tables.jl implementation instead of wrapping all the implementations into one.

If we want a unified interface for the table file types, we could have `TableDataset(path::String) = ...` which checks the file extension then invokes `CSVDataset` etc.

lorenzoh:
> If we want a unified interface for the table file types, we could have TableDataset(path::String) = ... which checks the file extension then invokes CSVDataset etc.

I like this idea, this way we can specialize on the format while having a uniform API. I would be in favor of consistently using `AbstractPath` types from FilePathsBase.jl, though. Maybe we can reexport the `p""` macro, then you can call `TableDataset(p"data.csv")`. Having a canonical format for paths avoids a lot of fiddly errors.

lorenzoh:
Just throwing out another possibility: Having one `TableDataset{T}` type that wraps a table of type `T` and then having `getobs(td::TableDataset) = getobstable(td.table)` and `nobs(td::TableDataset) = nobstable(td.table)` where we own `getobstable` and `nobstable` and can specialize them to the table type.

The advantage here is that we can then provide a default (perhaps slow) implementation for all tables. Also endusers will only deal with one type, i.e. calling `TableDataset` always returns a `TableDataset`.

darsnack:
Why not just do `getobs(td::TableDataset{<:DataFrame}, i)`? @manikyabard can you implement this? In addition to the code that you already have, you would write `getobs(td::TableDataset{<:DataFrame}, i)` and `getobs(td::TableData{<:CSV.File}, i)` (which are faster implementations that assume that `td.table` is a `DataFrame` or `CSV.File`).

darsnack:
Also, let's use `AbstractPath` too

lorenzoh:
Ah yes, that is functionally equivalent and more elegant 👍

src/datasets/containers.jl
```diff
@@ -45,5 +43,44 @@ isimagefile(file::File) = isimagefile(file.name)
...
+    else error("The Tables.jl implementation used should have either rowaccess or columnaccess.")
+    end
+end
+
+LearnBase.getobs(dataset::TableDataset{<:DataFrame}, idx) = dataset.table[idx, :]
+LearnBase.nobs(dataset::TableDataset{<:DataFrame}) = nrow(dataset.table)
 
-## TODO: TableDataset
+LearnBase.getobs(dataset::TableDataset{<:CSV.File}, idx) = dataset.table[idx]
+LearnBase.getobs(dataset::TableDataset{<:CSV.File}) = length(dataset.table)
```
darsnack, line 86:
```suggestion
LearnBase.nobs(dataset::TableDataset{<:CSV.File}) = length(dataset.table)

```

src/datasets/containers.jl
```diff
@@ -45,5 +43,44 @@ isimagefile(file::File) = isimagefile(file.name)
...
 
+#TableDataset
+
+struct TableDataset{T}
+    table::T #Should implement Tables.jl interface
+    TableDataset{T}(table::T) where T = Tables.istable(table) ? new{T}(table) : error("Object doesn't implement Tables.jl interface")
+end
+
+TableDataset(path::AbstractPath) = TableDataset(DataFrame(CSV.File(path)))
+TableDataset(table::T) where {T} = TableDataset{T}(table)
```
darsnack, line 54:
```suggestion
```

darsnack (changes requested):
Needs tests

src/datasets/containers.jl
```diff
@@ -45,5 +43,43 @@ isimagefile(file::File) = isimagefile(file.name)
...
+        return rowvals
+    else error("The Tables.jl implementation used should have either rowaccess or columnaccess.")
+    end
+end
+
+function LearnBase.nobs(dataset::TableDataset{T}) where {T}
+    if Tables.columnaccess(dataset.table)
+        return length(Tables.getcolumn(dataset.table, 1))
+    elseif Tables.rowaccess(dataset.table)
+        return length(Tables.rows(dataset.table)) # Lenght might not be defined, but has to be for this to work.
```
lorenzoh, line 76:
```suggestion
        return length(Tables.rows(dataset.table)) # length might not be defined, but has to be for this to work.
```

manikyabard:
I have added a few tests for TableDataset in `test/datasets/containers.jl`.

src/datasets/containers.jl
```diff
@@ -78,8 +79,8 @@ function LearnBase.nobs(dataset::TableDataset{T}) where {T}
     end
 end
 
-LearnBase.getobs(dataset::TableDataset{<:DataFrame}, idx) = dataset.table[idx, :]
+LearnBase.getobs(dataset::TableDataset{<:DataFrame}, idx) = [data for data in dataset.table[idx, :]]
```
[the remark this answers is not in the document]
manikyabard: Yeah for `DataFrame` this would make more sense. But is it fine to keep this return behaviour for `getobs(dataset::TableDataset{<:CSV.File}, idx)` and the general `getobs(dataset::TableDataset{T}, idx)` methods considering that the types could be immutable, and they may not not have `setindex` defined (as in the case of `CSV.Row`) which might make transformations difficult if we use a copy of the input directly?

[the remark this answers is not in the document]
ToucheSir: I think any transforms shouldn't be calling `setindex!` anyhow without a defensive copy first, so we should stick with `dataset.table[idx, :]` everywhere.

src/datasets/containers.jl
```diff
@@ -50,13 +50,14 @@ struct TableDataset{T}
...
 
+TableDataset(table::T) where {T} = TableDataset{T}(table)
 TableDataset(path::AbstractPath) = TableDataset(DataFrame(CSV.File(path)))
 
 function LearnBase.getobs(dataset::FastAI.Datasets.TableDataset{T}, idx) where {T}
     if Tables.rowaccess(dataset.table)
         for (index, row) in enumerate(Tables.rows(dataset.table))
             if index==idx
-                return row
+                return [data for data in row]
```
ToucheSir, line 60: Similar comment here about keeping the row object instead of converting it.

src/datasets/containers.jl
```diff
@@ -50,13 +50,14 @@ struct TableDataset{T}
     TableDataset{T}(table::T) where T = Tables.istable(table) ? new{T}(table) : error("Object doesn't implement Tables.jl interface")
 end
 
+TableDataset(table::T) where {T} = TableDataset{T}(table)
 TableDataset(path::AbstractPath) = TableDataset(DataFrame(CSV.File(path)))
 
 function LearnBase.getobs(dataset::FastAI.Datasets.TableDataset{T}, idx) where {T}
     if Tables.rowaccess(dataset.table)
         for (index, row) in enumerate(Tables.rows(dataset.table))
```
ToucheSir, line 58: It's not faster, but instead of a loop + index check you could write something like `first(Iterators.drop(Tables.rows(...), idx - 1))` (skip up until the index and return the next element, which will be the element at the index).

src/datasets/containers.jl
manikyabard: So the main concern I have is that, if we use an object returned by `getobs` for transformation, and let's say normalization is defined as
```julia
struct NormalizeRow <: DataAugmentation.Transform
	normstats
	normidxs
end

function DataAugmentation.apply(tfm::NormalizeRow, item::TabularItem; randstate=nothing)
	x = copy(item.data)
	for idx in tfm.normidxs
		colmean, colstd = tfm.normstats[idx]
		x[idx] = (x[idx] - colmean)/colstd
	end
	TabularItem(x)
end
```
There's a possibility that functions like `copy` or `setindex` won't be defined for the object returned (example being `CSV.Row` as I mentioned before). Maybe we can extract the data from the returned object and put it in a vector in the transformation step itself, but then the indexes won't really work with it. Is there a way to retain the original type from `getobs` and still make these kinds of transformation work?

src/datasets/containers.jl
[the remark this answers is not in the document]
lorenzoh: Agreeing with Brian that the transforms shouldn't mutate the arguments by default; we have `encode!`/`run!` for the inplace versions.

src/datasets/containers.jl
manikyabard: Yeah, I can do that.

test/datasets/containers.jl
```diff
@@ -0,0 +1,53 @@
...
+    @testset ExtendedTestSet "TableDataset from DataFrames" begin
+        testtable = DataFrame(
+            col1=[1, 2, 3, 4, 5], 
+            col2=["a", "b", "c", "d", "e"], 
+            col3=[10, 20, 30, 40, 50], 
+            col4=["A", "B", "C", "D", "E"],
+            col5=[100., 200., 300., 400., 500.],
+            split=["train", "train", "train", "valid", "valid"]
+        )
+        @inferred TableDataset TableDataset(testtable)
```
darsnack, line 36: What are you trying to test here? Maybe this should be
```julia
@test TableDataset(testtable) isa TableDataset{<:DataFrame}
```

test/datasets/containers.jl
```diff
@@ -0,0 +1,53 @@
...
+        td = TableDataset(testtable)
+
+        @test getobs(td, 1) == [1, "a", 10, "A", 100., "train"]
+        @test nobs(td) == 5 
+    end
+
+    @testset ExtendedTestSet "TableDataset from CSV" begin
+        path = datasetpath("adult_sample")
+        testtable = CSV.File(joinpath(path, "adult.csv"))
+        @inferred TableDataset TableDataset(testtable)
```
darsnack, line 46: Same here

src/datasets/containers.jl
darsnack: Is this necessary because of the inner constructor?

src/datasets/containers.jl
darsnack: Let's solve this within the transforms portion of the pipeline instead of forcing a vector here.

src/datasets/containers.jl
```diff
@@ -45,5 +43,44 @@ isimagefile(file::File) = isimagefile(file.name)
...
+    else error("The Tables.jl implementation used should have either rowaccess or columnaccess.")
+    end
+end
+
+function LearnBase.nobs(dataset::TableDataset{T}) where {T}
+    if Tables.columnaccess(dataset.table)
+        return length(Tables.getcolumn(dataset.table, 1))
+    elseif Tables.rowaccess(dataset.table)
+        return length(Tables.rows(dataset.table)) # length might not be defined, but has to be for this to work.
+    else error("The Tables.jl implementation used should have either rowaccess or columnaccess.")
```
darsnack, line 78:
```suggestion
    else
      error("The Tables.jl implementation used should have either rowaccess or columnaccess.")
```

src/datasets/containers.jl
```diff
@@ -45,5 +43,44 @@ isimagefile(file::File) = isimagefile(file.name)
...
+        rowvals = []
+        for i in 1:length(Tables.columnnames(dataset.table))
+            push!(rowvals, Tables.getcolumn(dataset.table, i)[idx])
+        end
+        return rowvals
+    else error("The Tables.jl implementation used should have either rowaccess or columnaccess.")
+    end
+end
+
+function LearnBase.nobs(dataset::TableDataset{T}) where {T}
```
darsnack, line 73:
```suggestion
function LearnBase.nobs(dataset::TableDataset)
```

src/datasets/containers.jl
```diff
@@ -45,5 +43,44 @@ isimagefile(file::File) = isimagefile(file.name)
...
+
+struct TableDataset{T}
+    table::T #Should implement Tables.jl interface
+    TableDataset{T}(table::T) where T = Tables.istable(table) ? new{T}(table) : error("Object doesn't implement Tables.jl interface")
+end
+
+TableDataset(table::T) where {T} = TableDataset{T}(table)
+TableDataset(path::AbstractPath) = TableDataset(DataFrame(CSV.File(path)))
+
+function LearnBase.getobs(dataset::FastAI.Datasets.TableDataset{T}, idx) where {T}
```
darsnack, line 56:
```suggestion
function LearnBase.getobs(dataset::FastAI.Datasets.TableDataset, idx)
```

src/datasets/containers.jl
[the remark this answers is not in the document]
darsnack: +1 for the existing suggestions

src/datasets/containers.jl
```diff
@@ -45,5 +43,44 @@ isimagefile(file::File) = isimagefile(file.name)
...
+        return length(Tables.rows(dataset.table)) # length might not be defined, but has to be for this to work.
+    else error("The Tables.jl implementation used should have either rowaccess or columnaccess.")
+    end
+end
+
+LearnBase.getobs(dataset::TableDataset{<:DataFrame}, idx) = [data for data in dataset.table[idx, :]]
+LearnBase.nobs(dataset::TableDataset{<:DataFrame}) = nrow(dataset.table)
 
-## TODO: TableDataset
+LearnBase.getobs(dataset::TableDataset{<:CSV.File}, idx) = [data for data in dataset.table[idx]]
```
darsnack, line 85: Same here as the comment on `getobs(::TableDataset{<:DataFrame}, idx)`.

darsnack (changes requested):
I left some comments. @ToucheSir and @lorenzoh covered the main concerns already.

Also, it looks like something has gone very wrong with a git rebase here. If it isn't immediately obvious how to solve it, I would make a local copy of `src/datasets/*` and the tests, then start with a fresh branch/PR and re-apply the changes.

test/datasets/containers.jl
manikyabard: Yeah this makes more sense. I have updated the tests accordingly.

test/Project.toml
```diff
@@ -1,9 +1,15 @@
...
+CSV = "336ed68f-0bac-5ca0-87d4-7b16caf5d00b"
 Colors = "5ae59095-9a9b-59fe-a467-6f913c188581"
 DLPipelines = "e6530d7c-7faa-4ede-a0d6-9eff9baad396"
 DataAugmentation = "88a5189c-e7ff-4f85-ac6b-e6158070f02e"
+DataFrames = "a93c6f00-e57d-5684-b7b6-d8193f3e46c0"
 FilePathsBase = "48062228-2e41-5def-b9a4-89aafe57970f"
 Flux = "587475ba-b771-5e3f-ad9e-33799f191a9c"
+JLD2 = "033835bb-8acc-5ee8-8aae-3f567f8a3819"
+Makie = "ee78f7c6-11fb-53f2-987a-cfe4a2b5a57a"
+ShowCases = "605ecd9f-84a6-4c9e-81e2-4798472b76a3"
```
darsnack, line 11: These seem orthogonal, maybe @lorenzoh can double-check?

test/datasets/containers.jl
```diff
@@ -0,0 +1,53 @@
...
+
+    @testset ExtendedTestSet "TableDataset from columnaccess table" begin
+        Tables.columnaccess(::Type{<:Tables.MatrixTable}) = true
+        Tables.rowaccess(::Type{<:Tables.MatrixTable}) = false
+        
+        testtable = Tables.table([1 4.0 "7"; 2 5.0 "8"; 3 6.0 "9"])
+        td = TableDataset(testtable)
+
+        @test [data for data in getobs(td, 2)] == [2, 5.0, "8"]
+        @test nobs(td) == 3
```
darsnack, line 24:
```suggestion
        @test nobs(td) == 3

        @test getobs(td, 1) isa NamedTuple
```

test/datasets/containers.jl
```diff
@@ -0,0 +1,53 @@
...
+@testset ExtendedTestSet "TableDataset" begin
+
+    @testset ExtendedTestSet "TableDataset from rowaccess table" begin
+        Tables.columnaccess(::Type{<:Tables.MatrixTable}) = false
+        Tables.rowaccess(::Type{<:Tables.MatrixTable}) = true
+        
+        testtable = Tables.table([1 4.0 "7"; 2 5.0 "8"; 3 6.0 "9"])
+        td = TableDataset(testtable)
+
+        @test [data for data in getobs(td, 1)] == [1, 4.0, "7"]
```
darsnack, line 12: Not sure if broadcasting works here:
```suggestion
        @test all(getobs(td, 1) .== [1, 4.0, "7"])
```

test/datasets/containers.jl
```diff
@@ -0,0 +1,53 @@
...
+        testtable = DataFrame(
+            col1=[1, 2, 3, 4, 5], 
+            col2=["a", "b", "c", "d", "e"], 
+            col3=[10, 20, 30, 40, 50], 
+            col4=["A", "B", "C", "D", "E"],
+            col5=[100., 200., 300., 400., 500.],
+            split=["train", "train", "train", "valid", "valid"]
+        )
+        @test TableDataset(testtable) isa TableDataset{<:DataFrame}
+        td = TableDataset(testtable)
```
darsnack, line 37:
```suggestion
        td = TableDataset(testtable)
        @test td isa TableDataset{<:DataFrame}
```

test/datasets/containers.jl
```diff
@@ -0,0 +1,53 @@
...
+
+        @test [data for data in getobs(td, 1)] == [1, "a", 10, "A", 100., "train"]
+        @test nobs(td) == 5 
+    end
+
+    @testset ExtendedTestSet "TableDataset from CSV" begin
+        path = datasetpath("adult_sample")
+        testtable = CSV.File(joinpath(path, "adult.csv"))
+        @test TableDataset(testtable) isa TableDataset{<:CSV.File}
+        td = TableDataset(testtable)
```
darsnack, line 47:
```suggestion
        td = TableDataset(testtable)
        @test td isa TableDataset{<:CSV.File}
```

test/datasets/containers.jl
```diff
@@ -0,0 +1,53 @@
...
+
+        @test [data for data in getobs(td, 2)] == [44, " Private", 236746, " Masters", 14.0, " Divorced", " Exec-managerial", " Not-in-family", " White", " Male", 10520, 0, 45, " United-States", ">=50k"]
```
darsnack, line 49:
```suggestion
        @test [data for data in getobs(td, 2)] == [44,
                                                   " Private",
                                                   236746,
                                                   " Masters",
                                                   14.0,
                                                   " Divorced",
                                                   " Exec-managerial",
                                                   " Not-in-family",
                                                   " White",
                                                   " Male",
                                                   10520,
                                                   0,
                                                   45,
                                                   " United-States",
                                                   ">=50k"]
```

darsnack (changes requested):
Seems you figured out the git issue. The PR looks good to me. I made some minor suggestions on the test cases.

test/datasets/containers.jl
manikyabard: It should work for `MatrixRow` objects.

darsnack (approved):
Pending a check from @lorenzoh on the packages, I think this is GTM.

test/Project.toml
lorenzoh: JLD2, Makie and ShowCases are already in the master branch, not sure why they're shown here. Might just be that the list was sorted on a ]pkg command.

lorenzoh (approved):
Since https://github.com/FluxML/ML-Coordination-Tracker/issues/34 states that the points, include documentation, we shouldn't forget about that. Though I suppose this as a basis for other work can be merged as is. We should definitely add more comprehensive documentation later.

lorenzoh:
@manikyabard The changes look good to me.  Can you remove the `Manifest.toml` from your forked repo? It seems to be causing the CI errors. If they run fine, we can merge this.

manikyabard:
> @manikyabard The changes look good to me. Can you remove the `Manifest.toml` from your forked repo? It seems to be causing the CI errors. If they run fine, we can merge this.

Okay, I deleted `Manifest.toml` from the branch.

lorenzoh:
Tests are failing. Did you run them locally? I think the problem is `include("imports.jl")` in `test/datasets/cotainers.jl` which should be `include("../imports.jl")` since it is in a subdirectory.

manikyabard:
> Tests are failing. Did you run them locally? I think the problem is `include("imports.jl")` in `test/datasets/cotainers.jl` which should be `include("../imports.jl")` since it is in a subdirectory.

So the path error should be fixed now, but it looks like there is still an error while downloading data in the testcase `TableDataset from CSV`, although the tests pass on my local system.

lorenzoh:
Could you remove the commented-out lines in the test case?

manikyabard:
Sure I'll do that.

lorenzoh:
Looks good, merging!

PR #28, opened 2021-04-08
word and character level word tokenizer.

### LINKED ISSUE #24: Add character and word level tokenizers
Add simple character and word level tokenizers that conform to the LearnBase.jl `getobs`/`nobs` interfaces.

SamuelzXu:
Some miscellaneous changes with the tokenizer being the main focus. Implements the LearnBase getobs and nobs methods. Uses the WordTokenizers module.

Closes #24

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ As an example, training an image classification model from scratch is as simple 
 using FastAI
 path = datasetpath("imagenette2-160")
 data = loadtaskdata(path, ImageClassificationTask)
-method = ImageClassification(Datasets.loadclassesclassification("imagenette2-160"), (160, 160))
+method = ImageClassification(Datasets.getclassesclassification("imagenette2-160"), (160, 160))
 learner = methodlearner(method, data, Models.xresnet18(), ToGPU(), Metrics(accuracy))
 fitonecycle!(learner, 5)
 ```
--- a/src/datasets/transformations.jl
+++ b/src/datasets/transformations.jl
@@ -41,8 +41,9 @@ struct NamedTupleData{TData, F}
     namedfs::NamedTuple{F}
 end
 
-LearnBase.nobs(data::NamedTupleData) = nobs(getfield(data, :data))
+# LearnBase functions {nobs, getobs} for NamedTupleData
 
+LearnBase.nobs(data::NamedTupleData) = nobs(getfield(data, :data))
 function LearnBase.getobs(data::NamedTupleData{TData, F}, idx::Int) where {TData, F}
     obs = getobs(getfield(data, :data), idx)
     namedfs = getfield(data, :namedfs)
@@ -132,6 +133,8 @@ struct JoinedData{T,N}
 end
 
 JoinedData(datas) = JoinedData(datas, nobs.(datas))
+
+# LearnBase functions {nobs, getobs} for JoinedData
 
 LearnBase.nobs(data::JoinedData) = sum(data.ns)
 function LearnBase.getobs(data::JoinedData, idx)
--- a/src/datasets/transformations.jl
+++ b/src/datasets/transformations.jl
@@ -1,4 +1,4 @@
-using WordTokenizers
+using WordTokenizers: TokenBuffer, flush!, character, isdone, spaces
 
 # mapobs
 
--- a/src/datasets/transformations.jl
+++ b/src/datasets/transformations.jl
@@ -171,9 +171,7 @@ function tokenize(type::Symbol,input)::AbstractArray{AbstractString}
     ts = TokenBuffer(input)
     if type === :chars
         while !isdone(ts)
-            while !isdone(ts)
-                character(ts)
-            end
+            character(ts)
         end
     elseif type === :words
         while !isdone(ts)
--- a/src/datasets/transformations.jl
+++ b/src/datasets/transformations.jl
@@ -157,7 +157,6 @@ getobs(jdata, 15) == 15
 """
 joinobs(datas...) = JoinedData(datas)
 
-
 # TODO: NamedTupleData transformation
 #
 # mdata = mapobs(data, (col1 = f1, col2 = f2))
--- a/src/datasets/transformations.jl
+++ b/src/datasets/transformations.jl
@@ -172,6 +172,7 @@ function tokenize(type::Symbol,input)::AbstractArray{AbstractString}
     if type === :chars
         while !isdone(ts)
             character(ts)
+            flush!(ts)
         end
     elseif type === :words
         while !isdone(ts)
--- a/src/datasets/transformations.jl
+++ b/src/datasets/transformations.jl
@@ -1,3 +1,4 @@
+using WordTokenizers
 
 # mapobs
 
@@ -159,6 +160,28 @@ getobs(jdata, 15) == 15
 ```
 """
 joinobs(datas...) = JoinedData(datas)
+
+"""
+    tokenize(type, input)
+
+Tokenizes an input string or stream into pieces depending on selected type. 
+"""
+
+function tokenize(type::Symbol,input)::AbstractArray{AbstractString}
+    ts = TokenBuffer(input)
+    if type === :chars
+        while !isdone(ts)
+            while !isdone(ts)
+                character(ts)
+            end
+        end
+    elseif type === :words
+        while !isdone(ts)
+            spaces(ts) || character(ts)
+        end
+    end
+    return ts.tokens
+end
 
 # TODO: NamedTupleData transformation
 #
--- a/src/datasets/transformations.jl
+++ b/src/datasets/transformations.jl
@@ -183,6 +183,13 @@ function tokenize(type::Symbol,input)::AbstractArray{AbstractString}
     return ts.tokens
 end
 
+# LearnBase functions {nobs, getobs} for AbstractArray{AbstractString}
+
+LearnBase.nobs(data::AbstractArray{AbstractString}) = size(data,1)
+function LearnBase.getobs(data::AbstractArray{AbstractString}, idx::Int)
+    return data[idx]
+end
+
 # TODO: NamedTupleData transformation
 #
 # mdata = mapobs(data, (col1 = f1, col2 = f2))
--- a/test/datasets/transformations.jl
+++ b/test/datasets/transformations.jl
@@ -37,10 +37,10 @@ include("../imports.jl")
     @testset ExtendedTestSet "tokenize" begin
         data1 = "The quick rabbit jumps over the lazy fox."
         data2 = " The quick rabbit jumps over the lazy fox."
-        tdata1 = tokenize(:words,data1)
-        tdata2 = tokenize(:words,data2)
-        tdata3 = tokenize(:chars,data1)
-        tdata4 = tokenize(:chars,data2)
+        tdata1 = Datasets.tokenize(:words,data1)
+        tdata2 = Datasets.tokenize(:words,data2)
+        tdata3 = Datasets.tokenize(:chars,data1)
+        tdata4 = Datasets.tokenize(:chars,data2)
         @test getobs(tdata1,3) == "rabbit"
         @test getobs(tdata2,3) == "rabbit"
         @test getobs(tdata3,3) == "e"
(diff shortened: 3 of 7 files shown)


src/datasets/transformations.jl
```diff
@@ -157,6 +161,33 @@ getobs(jdata, 15) == 15
 """
 joinobs(datas...) = JoinedData(datas)
 
+"""
+    tokenize(type, input)
+
+Tokenizes an input string or stream into pieces depending on selected type. 
+"""
```
ToucheSir, line 168: Is there any advantage to this over the default `WordTokenizers.tokenize`? I think it's fine to have a separate path for characters, but for words we'd prefer to rely on their implementation. One way to accomodate this may be to point users towards the experimental [split API](https://github.com/JuliaText/WordTokenizers.jl#experimental-api) and simply define an function overload which uses your character tokenizer.

src/datasets/transformations.jl
```diff
@@ -157,6 +161,33 @@ getobs(jdata, 15) == 15
...
+
+function tokenize(type::Symbol,input)::AbstractArray{AbstractString}
```
ToucheSir, line 170: The type declarations here can be safely omitted.

src/datasets/transformations.jl
```diff
@@ -157,6 +161,33 @@ getobs(jdata, 15) == 15
...
+        while !isdone(ts)
+            spaces(ts) || character(ts)
+        end
+    end
+    return ts.tokens
+end
+
+# LearnBase functions {nobs, getobs} for AbstractArray{AbstractString}
+
+LearnBase.nobs(data::AbstractArray{AbstractString}) = size(data,1)
```
ToucheSir, line 187: `length(data)` does the same thing as `size(data, 1)` here. We may want to look into a smarter data container eventually as special-casing a certain array type is non-optimal, but that shouldn't hold up this PR.

test/datasets/transformations.jl
```diff
@@ -33,4 +33,18 @@ include("../imports.jl")
...
+        data2 = " The quick rabbit jumps over the lazy fox."
+        tdata1 = Datasets.tokenize(:words,data1)
+        tdata2 = Datasets.tokenize(:words,data2)
+        tdata3 = Datasets.tokenize(:chars,data1)
+        tdata4 = Datasets.tokenize(:chars,data2)
+        @test getobs(tdata1,3) == "rabbit"
+        @test getobs(tdata2,3) == "rabbit"
+        @test getobs(tdata3,3) == "e"
+        @test getobs(tdata4,3) == "h"
+        #TODO: add stream tests
```
ToucheSir, line 48: By "stream tests", do you mean enumerating more than one token? If so, I agree those should be added.

README.md
```diff
@@ -10,7 +10,7 @@ As an example, training an image classification model from scratch is as simple
 using FastAI
 path = datasetpath("imagenette2-160")
 data = loadtaskdata(path, ImageClassificationTask)
-method = ImageClassification(Datasets.loadclassesclassification("imagenette2-160"), (160, 160))
+method = ImageClassification(Datasets.getclassesclassification("imagenette2-160"), (160, 160))
```
ToucheSir, line 13: This and the change below appear unrelated to the PR. Can you revert them?

ToucheSir (changes requested):
Thanks for the PR! I've left some comments and considerations, but it shouldn't take much to get this landed :)

test/datasets/transformations.jl
SamuelzXu: No, I meant to take a stream as input, which I have yet to figure out how to create as a test. By enumerate more than one token, do you mean check more than one index?

README.md
SamuelzXu: Sure, but they do cause bugs (namely, loadclasses... gave undefined function) during the tutorial.

src/datasets/transformations.jl
SamuelzXu: The advantage was that I can control the parameters, so that if WordTokenizers is updated we're not pointing to a function that doesn't exist anymore. I don't quite understand what you mean by the last sentence. Are you suggesting to not implement the tokenizer, and instead tell the user to use the WordTokenizer module's split functions, to implement the same thing with WordTokenizer.tokenize() in place of the case for ``:words``, or something else? The confusion is what you mean by "point".

README.md
ToucheSir: In that case it's better to open another quick PR. Generally if it's not blocking CI, then it's better to not put unrelated changes in.

src/datasets/transformations.jl
ToucheSir: That's correct. We can always re-export `tokenize` from `WordTokenizers` if need be. I mention the split API because it uses multiple dispatch to select which tokenizer to use and thus doesn't require a conditional over a symbol parameter like `:words` or `:chars`. This allows one to swap out tokenizer functions (see https://github.com/JuliaText/WordTokenizers.jl/blob/master/src/split_api.jl for the built-in list), so doing the equivalent of `:chars` could be as easy as calling `collect` on the input string.

README.md
lorenzoh: A PR with just that change would be appreciated, so it can be fixed swiftly 👍

src/datasets/transformations.jl
lorenzoh: Ah, also we shouldn't be implementing `getobs` for `Base` types like `AbstractArray` and `AbstractString`.

darsnack (changes requested):
To match what @lorenzoh suggested, we don't want to be defining `getobs` on generic types like `AbstractArray{<:AbstractString}`.

In addition to the existing suggestions, I think what we want is something like
```julia
struct Tokenizer{T, ...}
    data::T
    # ...
end

function LearnBase.getobs(tokenizer::Tokenizer, i)
    # fetch a new sample from data if necessary
    sample = getobs(tokenizer.data, ??)

    # tokenize the sample
    tokens = tokenize(sample)

    # return correct token
    return token[??]
end
```

This is pseudocode, but the idea is that the `Tokenizer` wraps `data` which is any stream of text. This could be a string or text dataset on disk wrapped in a lazy loader that read line by line. The `Tokenizer` is responsible for doing the index logic to know when it needs to fetch more text from `data`, tokenizing that text, and buffering the excess tokens. Certain calls to `getobs(::Tokenizer, ...)` will not require fetching more text from `data` but just returning what's already in the buffer.

ToucheSir:
As cool as incrementally reading text would be, I'm not sure it's feasible since you'd need to tokenize the entire stream to implement `nobs`. Hence the question becomes how to represent a collection of tokenized text samples rather than an individual one. Here it's worth looking at how [fastai](https://docs.fast.ai/text.data.html), [spaCy](https://spacy.io/usage/spacy-101) and more handle this.

darsnack:
True, I didn't think about that. At the very least, we should tokenize the full stream, store the tokens in `Tokenizer`, then return those on `getobs(::Tokenizer, i)`. As opposed to tokenizing the data into a generic vector of tokens and writing `getobs(::AbstractVector{<:AbstractString}, i)`.

SamuelzXu:
Ah, I see. I thought about this, and my logic going in was that since there was only one element in the struct, and we're working in and NLP context, it would be simpler for a coder to use this version. Can you point me to a place where I can read more about these design choices? (Like not implementing methods for base types @lorenzoh ) It feels like there is a lot of context I'm missing.

SamuelzXu:
@darsnack How do I update the PR? I read online I only have to push to my branch, but it doesn't seem to be showing up.

SamuelzXu:
Oh, wow, I must be blind! I see it now.

SamuelzXu:
@darsnack @ToucheSir could you guys help me out with the importing and extending bug?

lorenzoh:
This section of the Julia documentation explains why type piracy (defining methods of functions you don't own for types you don't own) should be avoided:
https://docs.julialang.org/en/v1/manual/style-guide/#Avoid-type-piracy

src/datasets/transformations.jl
```diff
@@ -41,8 +42,9 @@ struct NamedTupleData{TData, F}
     namedfs::NamedTuple{F}
 end
 
-LearnBase.nobs(data::NamedTupleData) = nobs(getfield(data, :data))
+# LearnBase functions {nobs, getobs} for NamedTupleData
```
darsnack, line 45: These changes are not related to this PR. We like to separate changes into different PRs when possible. Can you move these to another PR?

src/datasets/transformations.jl
```diff
@@ -41,8 +42,9 @@ struct NamedTupleData{TData, F}
...
 
+LearnBase.nobs(data::NamedTupleData) = nobs(getfield(data, :data))
```
darsnack, line 47: Same here (move to another PR)

src/datasets/transformations.jl
```diff
@@ -133,6 +135,8 @@ end
 
 JoinedData(datas) = JoinedData(datas, nobs.(datas))
 
+# LearnBase functions {nobs, getobs} for JoinedData
+
```
darsnack, line 139: Same here (move to another PR)

src/datasets/transformations.jl
```diff
@@ -157,6 +161,36 @@ getobs(jdata, 15) == 15
 """
 joinobs(datas...) = JoinedData(datas)
 
+"""
+    tokenize(type, input)
+    type = :words or :chars
+
+Tokenizes an input string or stream into pieces depending on selected type. 
+"""
```
darsnack, line 169: This should go above the actual function definition.

src/datasets/transformations.jl
```diff
@@ -157,6 +161,36 @@ getobs(jdata, 15) == 15
...
+
+struct Tokenizer
+    data::AbstractArray{AbstractString}
+end
```
darsnack, line 173: For performance reasons, in Julia we use type parameters to get concrete typed fields:
```suggestion
struct Tokenizer{T<:AbstractArray{<:AbstractString}}
    data::T
end
```
Also, should this be a `AbstractVector` or `AbstractArray`?

src/datasets/transformations.jl
```diff
@@ -157,6 +161,36 @@ getobs(jdata, 15) == 15
...
+    end
+    return Tokenizer(ts.tokens)
+end
+
+# LearnBase functions {nobs, getobs} for Tokens
+
+LearnBase.nobs(toks::Tokenizer) = length(toks.data)
+function LearnBase.getobs(toks::Tokenizer, idx)
+    return toks.data[idx]
+end
```
darsnack, line 193:
```suggestion
LearnBase.getobs(toks::Tokenizer, idx) = toks.data[idx]
```
Just to make things cleaner

darsnack (changes requested):

darsnack:
You need to add WordTokenizer.jl to the `test/Project.toml`.

src/datasets/transformations.jl
SamuelzXu: Will do.

src/datasets/transformations.jl
SamuelzXu: I chose AbstractArray because AbstractVector <: AbstractArray

test/datasets/transformations.jl
```diff
@@ -33,4 +33,13 @@ include("../imports.jl")
         jdata = joinobs(data1, data2)
         @test getobs(jdata, 15) == 15
     end
+
+    @testset ExtendedTestSet "tokenize" begin
+        tdata1 = Datasets.Tokenizer(Datasets.tokenize(:words,"The quick rabbit jumps over the lazy fox."))
+        tdata2 = Datasets.Tokenizer(Datasets.tokenize(:chars,"The quick rabbit jumps over the lazy fox."))
```
darsnack, line 39: Doesn't `tokenize` return a `Tokenizer` already?

darsnack (commented):
Can you rebase?

SamuelzXu:
@darsnack

src/datasets/transformations.jl
darsnack: I think separating out these changes is still outstanding

src/datasets/Datasets.jl
```diff
@@ -152,11 +152,13 @@ export
     # primitive containers
     FileDataset,
     TableDataset,
+    Tokenizer,
 
     # utilities
     isimagefile,
     loadfile,
     filename,
+    tokenize,
```
darsnack, line 161:
```suggestion
    tokenize_input,
```

SamuelzXu: @darsnack got it.

PR #58, opened 2021-04-12
tutorial errors: changed {load->get}classesclassification

SamuelzXu:
loadclassesclassification doesn't exist, so let's replace it with getclassesclassification.


SamuelzXu:
@darsnack

darsnack (approved):

Issue #32, opened 2021-04-10
Improve projective data augmentation


lorenzoh:
Currently augmentation on images and keypoints is restricted to a random crop by default. fastai uses much more aggressive augmentation, and so should we. This should be straightforward to implement using DataAugmentation.jl's composable `ProjectiveTransform`.

lorenzoh:
These augmentations will be added to DataAugmentation.jl first and then used in FastAI.jl. 

- Random lighting (brightness and contrast adjustment) [DataAugmentation.jl#28](https://github.com/lorenzoh/DataAugmentation.jl/pull/28)

lorenzoh:
Implemented with the DataAugmentation.jl 0.2.2 release and `FastAI.augs_projection`, `FastAI.augs_color`. See [this tutorial](https://fluxml.ai/FastAI.jl/dev/docs/howto/augmentvision.md.html).

PR #67, opened 2021-04-18
Better vision augmentations

lorenzoh:
Resolves issue #32.


lorenzoh:
Tests will fail until DataAugmentation.jl#39 is released.

PR #116, opened 2021-04-28
Move to DataAugmentation v0.2.0

lorenzoh:


lorenzoh:
Will fail until v0.2.0 is registered

PR #124, opened 2021-06-13
Add tabular model

manikyabard:
Adds a tabular model based on the python fastai implementation.


manikyabard:
The embedding layer used here is based on https://github.com/FluxML/Flux.jl/pull/1516.

src/models/tabularmodel.jl
```diff
@@ -0,0 +1,79 @@
...
+
+function (tm::TabularModel)(x)
+	x_cat, x_cont = x
+	if tm.n_emb != 0
+		x = [e(x_cat[i, :]) for (i, e) in enumerate(tm.embeds)]
+		x = vcat(x...)
+		x = tm.emb_drop(x)
+	end
+	if tm.n_cont != 0
+		if (tm.bn_cont != false)
```
lorenzoh, line 73: Equivalent to `if tm.bn_cont == true` which is equivalent to `if tm.bn_cont`.
```suggestion
		if tm.bn_cont
```

src/models/tabularmodel.jl
```diff
@@ -0,0 +1,79 @@
...
+	end
+	embedslist = [Embedding(ni, nf) for (ni, nf) in emb_szs]
+	emb_drop = Dropout(embed_p)
+	bn_cont = bn_cont ? BatchNorm(n_cont) : false
+	n_emb = sum(size(embedlayer.weight)[1] for embedlayer in embedslist)
+	sizes = append!(zeros(0), [n_emb+n_cont], layers, [out_sz])
+	actns = append!([], [act_cls for i in 1:(length(sizes)-1)], [nothing])
+	_layers = [linbndrop(Int64(sizes[i]), Int64(sizes[i+1]), use_bn=(use_bn && ((i!=(length(actns)-1)) || bn_final)), p=p, act=a, lin_first=lin_first) for (i, (p, a)) in enumerate(zip(push!(ps, 0.), actns))]
+	if !isnothing(y_range)
+		push!(_layers, Chain(@. x->Flux.sigmoid(x) * (y_range[2] - y_range[1]) + y_range[1]))
```
ToucheSir, line 59: This doesn't need to be wrapped in a Chain. Also, is the sigmoid range meant to replace the activation of the last linear layer or to run after it?

src/models/tabularmodel.jl
```diff
@@ -0,0 +1,79 @@
...
+	layers = Chain(_layers...)
+	TabularModel(embedslist, emb_drop, bn_cont, n_emb, n_cont, layers)
+end
+
+function (tm::TabularModel)(x)
+	x_cat, x_cont = x
+	if tm.n_emb != 0
+		x = [e(x_cat[i, :]) for (i, e) in enumerate(tm.embeds)]
+		x = vcat(x...)
+		x = tm.emb_drop(x)
```
ToucheSir, line 70: I wonder if there's a way to use `Parallel` for `tm.embeds` instead of writing all this out manually?

src/models/tabularmodel.jl
```diff
@@ -0,0 +1,79 @@
...
+	if ps isa Number
+		ps = fill(ps, length(layers))
+	end
+	embedslist = [Embedding(ni, nf) for (ni, nf) in emb_szs]
+	emb_drop = Dropout(embed_p)
+	bn_cont = bn_cont ? BatchNorm(n_cont) : false
+	n_emb = sum(size(embedlayer.weight)[1] for embedlayer in embedslist)
+	sizes = append!(zeros(0), [n_emb+n_cont], layers, [out_sz])
+	actns = append!([], [act_cls for i in 1:(length(sizes)-1)], [nothing])
+	_layers = [linbndrop(Int64(sizes[i]), Int64(sizes[i+1]), use_bn=(use_bn && ((i!=(length(actns)-1)) || bn_final)), p=p, act=a, lin_first=lin_first) for (i, (p, a)) in enumerate(zip(push!(ps, 0.), actns))]
```
ToucheSir, line 57: this could benefit from splitting into a few lines. Also, do you recall why the `Int64` cast was added?

src/models/tabularmodel.jl
darsnack: You could do `Parallel(vcat, [Embedding(ni, nf) for (ni, nf) in emb_szs])`. In the `Chain`, before the embedding layers, you would need `x -> ntuple(i -> x[i, :], length(emb_szs))` to split a single `x` into multiple arguments.

ToucheSir: I think splatting `eachslice(x, dims=1)` would work too.

src/models/tabularmodel.jl
darsnack: I would echo this suggestion. In particular, instead of having complicated `use_bn` logic, just use two lines. First, iterate over the `1:(n - 1)` elements, then just do `push!(_layers, # last element version)`.

Also, I think this will be cleaner as a standard `for`-loop instead of a generator. Instead of pre-computing `sizes`, could we use `Flux.outputsize` inside the loop?

src/models/tabularmodel.jl
```diff
@@ -0,0 +1,79 @@
...
+	n_cont
+	layers
+end
+
+function TabularModel(
+		layers; 
+		emb_szs,
+		n_cont,
+		out_sz,
+		ps::Union{Tuple, Vector, Number, Nothing}=nothing,
```
darsnack, line 35: I would recommend pushing the `isnothing` etc. logic to `linbndrop` (it already handles the 0 case). It can be extended to handle `nothing` and `false` too.

src/models/tabularmodel.jl
```diff
@@ -0,0 +1,79 @@
...
+		act_cls=Flux.relu,
+		lin_first::Bool=true)
+
+	n_cont = Int64(n_cont)
+	if isnothing(ps)
+		ps = zeros(length(layers))
+	end
+	if ps isa Number
+		ps = fill(ps, length(layers))
+	end
```
darsnack, line 50: In addition to the comment about `isnothing(p)` being part of `linbndrop`, this can also be deleted by doing `Iterators.cycle(ps)` when building `_layers`.

src/models/tabularmodel.jl
```diff
@@ -0,0 +1,79 @@
...
+	embedslist = [Embedding(ni, nf) for (ni, nf) in emb_szs]
+	emb_drop = Dropout(embed_p)
+	bn_cont = bn_cont ? BatchNorm(n_cont) : false
```
darsnack, line 53: Could make this `bn_cont = bn_cont ? BatchNorm(n_cont) : identity` then avoid the if/else logic in the forward pass.

src/models/tabularmodel.jl
darsnack: In the case of `y_range == nothing`, should it be `sigmoid`? Right no there is no activation at all.

src/models/tabularmodel.jl
```diff
@@ -0,0 +1,79 @@
...
+	n_emb = sum(size(embedlayer.weight)[1] for embedlayer in embedslist)
+	sizes = append!(zeros(0), [n_emb+n_cont], layers, [out_sz])
+	actns = append!([], [act_cls for i in 1:(length(sizes)-1)], [nothing])
```
darsnack, line 56: This can be avoided if you split the final layer into a separate line.

src/models/tabularmodel.jl
```diff
@@ -0,0 +1,79 @@
...
+		x = [e(x_cat[i, :]) for (i, e) in enumerate(tm.embeds)]
+		x = vcat(x...)
+		x = tm.emb_drop(x)
+	end
+	if tm.n_cont != 0
+		if (tm.bn_cont != false)
+			x_cont = tm.bn_cont(x_cont)
+		end
+		x = tm.n_emb!=0 ? vcat(x, x_cont) : x_cont
+	end
```
darsnack, line 77: I would suggest making the whole categorical branch/continuous branch portion of the model into a `Parallel`. Something like `Parallel(vcat, embeds, bn_cont)`. It should gracefully handle the cases when either the categorical or continuous vectors are empty.

darsnack (commented):
This is a good start, but I think we can improve the readability on top of the original Python code a lot here. In particular, I think there are two places where `Parallel` will intuitively express the operation and eliminate a lot of boilerplate both in the constructor and the forward pass.

src/models/tabularmodel.jl
darsnack: If it should be `sigmoid`, then I think it makes more sense to force `y_range::Tuple` and make the default as `(0, 1)`. Just avoid `nothing` entirely.

src/models/tabularmodel.jl
manikyabard: Okay. I'm even considering removing `nothing` as it won't work directly with `Iterators.cycle` if we do use that for `ps`.

src/models/tabularmodel.jl
manikyabard: Yeah I'll try doing this.

src/models/tabularmodel.jl
manikyabard: I think the `Int64` cast was added because `BatchNorm` and `Dense` don't seem to work with floats. I could maybe restrict the types to `Int64` only though.

manikyabard: Yeah, I can create the last `linbndrop` outside the loop. Although I'm not sure how the sizes for `Flux.outputsize` would work with the `Parallel` layer before it. Would it be better than using `sizes`?

src/models/tabularmodel.jl
manikyabard: No, I don't think there is an activation if `y_range == nothing` according to the [python implementation](https://github.com/fastai/fastai/blob/01a4363580215440771df4d273ce34c9dd7e46a7/fastai/tabular/model.py#L41).

src/models/tabularmodel.jl
manikyabard: Yeah I considered using `Parallel` first but just wasn't sure how to handle the input for that. This should simplify things a lot

src/models/tabularmodel.jl
```diff
@@ -0,0 +1,53 @@
...
+	n_emb = sum(size(embedlayer.weight)[1] for embedlayer in embedslist)
+	sizes = append!(zeros(0), [n_emb+n_cont], layers)
+	actns = append!([], [act_cls for i in 1:(length(sizes)-1)])
+
+	_layers = []
+	for (i, (p, a)) in enumerate(zip(Iterators.cycle(ps), actns))
+		layer = linbndrop(Int64(sizes[i]), Int64(sizes[i+1]), use_bn=use_bn, p=p, act=a, lin_first=lin_first)
+		push!(_layers, layer)
+	end
+	push!(_layers, linbndrop(Int64(last(sizes)), Int64(out_sz), use_bn=bn_final, lin_first=lin_first))
```
darsnack, line 50:
```suggestion
        n_cat = first(Flux.outputsize(embeds, (length(emb_szs), 1)))
	classifiers = [
	    linbndrop(n_cat, first(layers); use_bn=use_bn, p=first(ps), act=act_cls, lin_first=lin_first),
	    [linbndrop(isize, osize; use_bn=use_bn, p=p, act=act_cls, lin_first=lin_first)
	      for (isize, osize, p) in zip(layers[1:(end - 2)], layers[2:(end - 1)], Iterators.cycle(Base.tail(ps)))]...,
	    linbndrop(last(layers)), out_sz; use_bn=bn_final, lin_first=lin_first)
	]
```

darsnack (commented):
from call

src/models/tabularmodel.jl
```diff
@@ -0,0 +1,53 @@
...
+
+	embedslist = [Embedding(ni, nf) for (ni, nf) in emb_szs]
+	emb_drop = Dropout(embed_p)
+	embeds = Chain(x -> ntuple(i -> x[i, :], length(emb_szs)), Parallel(vcat, embedslist...), emb_drop)
+
+	bn_cont = bn_cont ? BatchNorm(n_cont) : identity
+
+	n_emb = sum(size(embedlayer.weight)[1] for embedlayer in embedslist)
+	sizes = append!(zeros(0), [n_emb+n_cont], layers)
+	actns = append!([], [act_cls for i in 1:(length(sizes)-1)])
```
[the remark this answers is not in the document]
darsnack: Can delete `sizes`.

src/models/tabularmodel.jl
ToucheSir: I would go so far as to convert the comprehension into a plain for loop with `push!`. Not any less efficient than the splat and would help with line lengths.

darsnack: Yeah maybe that's better here since we are using so many iterables in the comprehension

src/models/tabularmodel.jl
```diff
@@ -0,0 +1,52 @@
+function emb_sz_rule(n_cat)
```
lorenzoh, line 1: Can we add a comment with a link to where this is taken from?

manikyabard: Sure I can add [this link](https://github.com/fastai/fastai/blob/2742fe844573d06e700f869839fb9ec5f3a9bca9/fastai/tabular/model.py#L12). I believe they got this formula experimentally.

src/models/tabularmodel.jl
manikyabard: Seeing that `Flux.outputsize` isn't working with `Embedding` layers, I was thinking of keeping the current way of calculating `n_emb` for now.

src/models/tabularmodel.jl
```diff
@@ -0,0 +1,56 @@
...
+        act_cls=Flux.relu,
+        lin_first::Bool=true)
+
+    embedslist = [Embedding(ni, nf) for (ni, nf) in emb_szs]
+    emb_drop = Dropout(embed_p)
+    embeds = Chain(x -> ntuple(i -> x[i, :], length(emb_szs)), Parallel(vcat, embedslist...), emb_drop)
+
+    bn_cont = bn_cont ? BatchNorm(n_cont) : identity
+
+    n_emb = sum(size(embedlayer.weight)[1] for embedlayer in embedslist)
```
darsnack, line 41: This would be clearer with `Flux.outputsize`

src/models/tabularmodel.jl
```diff
@@ -0,0 +1,56 @@
...
+
+    ps = Iterators.cycle(ps)
+    classifiers = []
+
+    first_ps, ps = Iterators.peel(ps)
+    push!(classifiers, linbndrop(n_emb+n_cont, first(layers); use_bn=use_bn, p=first_ps, lin_first=lin_first, act=act_cls))
+    for (isize, osize, p) in zip(layers[1:(end-1)], layers[2:(end)], ps)
```
darsnack, line 48:
```suggestion
    for (isize, osize, p) in zip(layers[1:(end-2)], layers[2:(end-1)], ps)
```
Shouldn't this be not including the last element in `layers`?

src/models/tabularmodel.jl
```diff
@@ -0,0 +1,56 @@
...
+        layer = linbndrop(isize, osize; use_bn=use_bn, p=p, act=act_cls, lin_first=lin_first)
+        push!(classifiers, layer)
+    end
+    push!(classifiers, linbndrop(last(layers), out_sz; use_bn=bn_final, lin_first=lin_first))
+
+    layers = isnothing(y_range) ? Chain(Parallel(vcat, embeds, bn_cont), classifiers...) : Chain(Parallel(vcat, embeds, bn_cont), classifiers..., @. x->Flux.sigmoid(x) * (y_range[2] - y_range[1]) + y_range[1])
```
darsnack, line 54: Let's switch this to something like `final_activation = identity` as a kwarg.

darsnack (commented):
Looking really good. Left a few comments.

src/models/tabularmodel.jl
manikyabard: I'm not completely sure that this will work because the last special case only goes from last element in `layers` to `out_sz`, and the layer for doing second last element to last element in `layers` would not be there. So if for eg. we had only 2 elements in `layers`, the loop probably won't be entered at all.

src/models/tabularmodel.jl
manikyabard: Okay, then I can make a `sigmoidrange` function and remove the `y_range` part. If users want to use this then something like `x -> sigmoidrange(x, 4, 6)` (for getting values between 4 and 6) should work.

src/models/tabularmodel.jl
darsnack: Never mind, yes you're right!

src/models/tabularmodel.jl
darsnack: Yup, that's what I was thinking too

src/models/tabularmodel.jl
darsnack: @manikyabard can you paste the error you got with `Embedding` and `outputsize`?

manikyabard: Sure, so if I do something like this
```julia
Flux.outputsize(Embedding(2, 3), (2,1))
```
the error I get is
```
ArgumentError: unable to check bounds for indices of type Flux.NilNumber.Nil

Stacktrace:
  [1] checkindex(#unused#::Type{Bool}, inds::Base.OneTo{Int64}, i::Flux.NilNumber.Nil)
    @ Base ./abstractarray.jl:671
  [2] checkindex
    @ ./abstractarray.jl:686 [inlined]
  [3] checkbounds_indices (repeats 2 times)
    @ ./abstractarray.jl:642 [inlined]
  [4] checkbounds
    @ ./abstractarray.jl:595 [inlined]
  [5] checkbounds
    @ ./abstractarray.jl:616 [inlined]
  [6] _getindex
    @ ./multidimensional.jl:831 [inlined]
  [7] getindex
    @ ./abstractarray.jl:1170 [inlined]
  [8] Embedding
    @ ~/.julia/packages/Flux/wii6E/src/layers/basic.jl:421 [inlined]
  [9] (::Embedding{Matrix{Float32}})(x::Matrix{Flux.NilNumber.Nil})
    @ Flux ~/.julia/packages/Flux/wii6E/src/layers/basic.jl:422
 [10] #outputsize#279
    @ ~/.julia/packages/Flux/wii6E/src/outputsize.jl:93 [inlined]
 [11] outputsize(m::Embedding{Matrix{Float32}}, inputsizes::Tuple{Int64, Int64})
    @ Flux ~/.julia/packages/Flux/wii6E/src/outputsize.jl:92
 [12] top-level scope
    @ In[213]:1
 [13] eval
    @ ./boot.jl:360 [inlined]
 [14] include_string(mapexpr::typeof(REPL.softscope), mod::Module, code::String, filename::String)
    @ Base ./loading.jl:1094
```

src/models/tabularmodel.jl
```diff
@@ -0,0 +1,65 @@
...
+
+    first_ps, ps = Iterators.peel(ps)
+    push!(classifiers, linbndrop(n_emb+n_cont, first(layers); use_bn=use_bn, p=first_ps, lin_first=lin_first, act=act_cls))
+    for (isize, osize, p) in zip(layers[1:(end-1)], layers[2:(end)], ps)
+        layer = linbndrop(isize, osize; use_bn=use_bn, p=p, act=act_cls, lin_first=lin_first)
+        push!(classifiers, layer)
+    end
+    push!(classifiers, linbndrop(last(layers), out_sz; use_bn=bn_final, lin_first=lin_first))
+    layers = Chain(
+        x -> tuple(x...),
```
darsnack, line 59:
```suggestion
```

src/models/tabularmodel.jl
```diff
@@ -0,0 +1,65 @@
...
+        Parallel(vcat, embeds, Chain(x -> ndims(x)==1 ? Flux.unsqueeze(x, 2) : x, bn_cont)),
```
darsnack, line 60: Move the `unsqueeze` to the encode method

darsnack (commented):
Just some comments from the call. Also we want to refactor as `TabularModel(EmbeddingBackbone, ContinuousBackbone, layers)` like we discussed.

lorenzoh:
Great to have tests! I think there're just some imports missing which is why they're failing

manikyabard:
> Great to have tests! I think there're just some imports missing which is why they're failing

Yeah I think it's probably because the `Embedding` PRs aren't merged yet.

lorenzoh:
Tests are still failing because `Embedding` PR is not released I think

src/models/tabularmodel.jl
```diff
@@ -0,0 +1,81 @@
...
+        lin_first=true)
+    ps = Iterators.cycle(ps)
+    classifiers = []
+
+    for (isize, osize, p) in zip(layers[1:(end-1)], layers[2:end], ps)
+        layer = linbndrop(isize, osize; use_bn=use_bn, p=p, act=act_cls, lin_first=lin_first)
+        push!(classifiers, layer)
+    end
+    Chain(classifiers...)
+end
```
darsnack, line 49: I wasn't thinking that the linear chain should be customizable. I think this loop should get pushed into `TabularModel`. My suggestion was that `bn_final` and `final_activation` be lumped into a single `classifier` argument to `TabularModel` which is positional and defaults to `Dense(in, out)`.

src/models/tabularmodel.jl
```diff
@@ -0,0 +1,81 @@
...
+    Chain(
+        x -> tuple(eachrow(x)...), 
+        Parallel(vcat, embedslist), 
+        emb_drop
+    )
+end
+
+function continuousbackbone(n_cont)
+    n_cont > 0 ? BatchNorm(n_cont) : identity
+end
```
darsnack, line 32: I don't know how useful it is to have this function.

src/models/tabularmodel.jl
```diff
@@ -0,0 +1,81 @@
...
+        final_activation
+    )
+end
+
+function TabularModel(
+        catcols,
+        n_cont::Number,
+        out_sz::Number,
+        layers=[200, 100];
+        catdict,
```
darsnack, line 69: I have some questions about understanding `catdict` that we can discuss during the call.

darsnack (commented):
Let's maybe knock out the final restructuring of this on the next call. I think it's possible to eliminate a lot of these keywords while maintaining flexibility (or at least restructure them so that it is clearer what each keyword does).

src/models/tabularmodel.jl
lorenzoh: Can you add it?

src/models/tabularmodel.jl
```diff
@@ -0,0 +1,129 @@
...
+    )
+end
+
+function tabular_continuous_backbone(n_cont)
+    BatchNorm(n_cont)
+end
+
+"""
+    TabularModel(catbackbone, contbackbone, [finalclassifier]; kwargs...)
+    TabularModel(catcols, `n_cont::Number, outsz::Number[; kwargs...)
```
lorenzoh, line 50:
```suggestion
    TabularModel(catcols, n_cont::Number, outsz::Number; kwargs...])
```

lorenzoh (changes requested):

darsnack (changes requested):
Mostly minor name changes to make things more readable. Only major change is to collapse `catcols` and `cardinalitydict` into a single argument. We want to optimize the API to be simple for the common case.

src/models/tabularmodel.jl
ToucheSir: Is that final `]` necessary either? Never seen that syntax for optional args before.

src/models/tabularmodel.jl
```diff
@@ -0,0 +1,129 @@
...
+for each column present in `cols`. 
+
+## Keyword arguments
+
+- `sz_dict`: An indexable collection which may contain the required embedding 
+    size for a particular column present in `cols`. If not passed, then the size is
+    calculated using fastai's rule of thumb for embedding dimensions.
+"""
+
+function get_emb_sz(cardinalitydict, cols; sz_dict=nothing)
```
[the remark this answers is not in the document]
manikyabard: Sure we can do this. Just to make it simpler for users, maybe another method which takes in a `Dict` to override easily would be helpful here.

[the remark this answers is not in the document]
darsnack: Yeah actually that makes more sense. One method with no size overrides that takes the cardinality vector as input. Another that takes `Dict`s like you have now.

src/models/tabularmodel.jl
```diff
@@ -0,0 +1,141 @@
...
+number of classes in categorical columns present in `cardinalities` and adjusting for nans. 
+
+## Keyword arguments
+
+- `size_overrides`: Depending on the method used, this could either be a collection of 
+    Integers and `nothing` or an indexable collection with column name as key and size
+    to override it with as the value. In the first case, the integer present at any index 
+    will be used to override the rule of thumb for getting embedding sizes.
+- `categorical_cols`: A collection of categorical column names.
+"""
```
darsnack, line 26: This docstring is better off split into two separate docstrings for each method. Both will show up automatically in the docs, but you will be able to tailor the explanation of `size_overrides` to each method (instead of explaining both in one paragraph).

src/models/tabularmodel.jl
```diff
@@ -0,0 +1,141 @@
...
+
+function get_emb_sz(cardinalities, size_overrides=fill(nothing, length(cardinalities)))
+    map(Iterators.enumerate(cardinalities)) do (i, cardinality)
+        emb_dim = isnothing(size_overrides[i]) ? emb_sz_rule(cardinality+1) : size_overrides[i]
+        (Int64(cardinality)+1, Int64(emb_dim))
+    end
+end
```
darsnack, line 33:
```suggestion
get_emb_sz(cardinalities::AbstractVector{<:Integer}, size_overrides=fill(nothing, length(cardinalities))) =
    map(zip(cardinalities, size_overrides)) do (cardinality, override)
        emb_dim = isnothing(override) ? emb_sz_rule(cardinality + 1) : Int64(override)
        return (cardinality + 1, emb_dim)
    end
```

src/models/tabularmodel.jl
```diff
@@ -0,0 +1,141 @@
...
+    keylist = keys(size_overrides)
+    overrides = map(catcols) do col
+        col in keylist ? size_overrides[col] : nothing
+    end
+    get_emb_sz(cardinalities, overrides)
+end
+
+function sigmoidrange(x, low, high)
+    @. Flux.sigmoid(x) * (high - low) + low
+end
```
darsnack, line 45:
```suggestion
sigmoidrange(x, low, high) = @. Flux.sigmoid(x) * (high - low) + low
```

src/models/tabularmodel.jl
```diff
@@ -0,0 +1,141 @@
...
+function emb_sz_rule(n_cat)
+    min(600, round(1.6 * n_cat^0.56))
+end
+
+"""
+    get_emb_sz(cardinalities, [size_overrides])
+    get_emb_sz(cardinalities; catcols, [size_overrides])
+
+Returns a collection of tuples containing embedding dimensions corresponding to 
+number of classes in categorical columns present in `cardinalities` and adjusting for nans.
```
darsnack, line 17:
```suggestion
number of classes in categorical columns present in `cardinalities` and adjusting for NaNs. 
```

src/models/tabularmodel.jl
```diff
@@ -0,0 +1,141 @@
...
+    )
+end
+
+function tabular_continuous_backbone(n_cont)
+    BatchNorm(n_cont)
+end
+
+"""
+    TabularModel(catbackbone, contbackbone, [finalclassifier]; kwargs...)
+    TabularModel(`n_cont::Number, outsize::Number[; kwargs...])
```
darsnack, line 63:
```suggestion
    TabularModel(n_cont, outsize [; kwargs...])
```

src/models/tabularmodel.jl
```diff
@@ -0,0 +1,141 @@
...
+
+Create a tabular model which takes in a tuple of categorical values 
+(label or one-hot encoded) and continuous values. The default categorical backbone is
+a Parallel of Embedding layers corresponding to each categorical variable, and continuous
+variables are just BatchNormed. The output from these backbones is then passed through
+a final classifier block.
```
darsnack, line 69: This is another case where two separate docstrings would help. `catbackbone`, etc. don't have default values except implicitly in second method.

src/models/tabularmodel.jl
```diff
@@ -0,0 +1,141 @@
...
+- `size_overrides`: An optional argument which corresponds to a collection containing 
+    embedding sizes to override the value returned by the "rule of thumb" for a particular index 
+    corresponding to `cardinalities`, or `nothing`.
+"""
+
+function TabularModel(
+        catbackbone, 
+        contbackbone;
+        outsize,
+        layersizes=[200, 100],
```
darsnack, line 93:
```suggestion
        layersizes=(200, 100),
```

src/models/tabularmodel.jl
```diff
@@ -0,0 +1,141 @@
...
+    
+    tabularbackbone = Parallel(vcat, catbackbone, contbackbone)
+
+    classifierin = mapreduce(layer -> size(layer.weight)[1], +, catbackbone[2].layers;
+                             init = contbackbone.chs)
+    dropout_rates = Iterators.cycle(dropout_rates)
+    classifiers = []
+
+    first_ps, dropout_rates = Iterators.peel(dropout_rates)
+    push!(classifiers, linbndrop(classifierin, first(layersizes); use_bn=batchnorm, p=first_ps, lin_first=linear_first, act=activation))
```
darsnack, line 116:
```suggestion
    push!(classifiers, linbndrop(classifierin, first(layersizes);
                                 use_bn=batchnorm, p=first_ps, lin_first=linear_first, act=activation))
```

src/models/tabularmodel.jl
```diff
@@ -0,0 +1,141 @@
...
+        tabularbackbone,
+        classifiers...,
+        finalclassifier
+    )
+end
+
+function TabularModel(
+        n_cont::Number,
+        outsize::Number,
+        layersizes=[200, 100];
```
darsnack, line 133:
```suggestion
        layersizes=(200, 100);
```

src/models/tabularmodel.jl
```diff
@@ -0,0 +1,141 @@
+"""
+    emb_sz_rule(n_cat)
+
+Returns an embedding size corresponding to the number of classes for a 
+categorical variable using the rule of thumb present in python fastai.
+(see https://github.com/fastai/fastai/blob/2742fe844573d06e700f869839fb9ec5f3a9bca9/fastai/tabular/model.py#L12)
+"""
+function emb_sz_rule(n_cat)
+    min(600, round(1.6 * n_cat^0.56))
+end
```
darsnack, line 10:
```suggestion
emb_sz_rule(n_cat) = min(600, round(Int, 1.6 * n_cat^0.56))
```

src/models/tabularmodel.jl
```diff
@@ -0,0 +1,141 @@
...
+
+function tabular_embedding_backbone(embedding_sizes, dropoutprob=0.)
+    embedslist = [Flux.Embedding(ni, nf) for (ni, nf) in embedding_sizes]
+    emb_drop = dropoutprob==0. ? identity : Dropout(dropoutprob)
+    Chain(
+        x -> tuple(eachrow(x)...), 
+        Parallel(vcat, embedslist), 
+        emb_drop
+    )
+end
```
darsnack, line 55:
```suggestion
function tabular_embedding_backbone(embedding_sizes, dropout_rates=0.)
    embedslist = [Flux.Embedding(ni, nf) for (ni, nf) in embedding_sizes]
    emb_drop = iszero(dropout_rates) ? identity : Dropout(dropout_rates)
    Chain(
        x -> tuple(eachrow(x)...), 
        Parallel(vcat, embedslist), 
        emb_drop
    )
end
```

src/models/tabularmodel.jl
```diff
@@ -0,0 +1,141 @@
...
+
+function tabular_continuous_backbone(n_cont)
+    BatchNorm(n_cont)
+end
```
darsnack, line 59:
```suggestion
tabular_continuous_backbone(n_cont) = BatchNorm(n_cont)
```

darsnack (changes requested):
Small nitpicks mostly. Let's also split some of these docstrings so it is less confusing trying to explain multiple sets of arguments in a single paragraph.

src/models/tabularmodel.jl
ToucheSir: `emb_drop` could even be further inlined if desired. If preserving model structure is more important than a bit of lost performance, then passing `active=iszero(dropout_rates)` instead of using a ternary is also a valid option.

manikyabard: Is `active` an argument to something? How would that work?

ToucheSir: `active` is the third arg to `BatchNorm`, yes. I wouldn't worry about that now though, it's a minor tweak we can always revisit later.

src/models/tabularmodel.jl
```diff
@@ -0,0 +1,158 @@
...
+- `categorical_cols`: A collection of categorical column names.
+"""
+
+function get_emb_sz(cardinalities::AbstractVector{<:Integer}, categorical_cols::Tuple, size_overrides=Dict())
+    keylist = keys(size_overrides)
+    overrides = collect(map(categorical_cols) do col
+        col in keylist ? size_overrides[col] : nothing
+    end)
+    get_emb_sz(cardinalities, overrides)
+end
```
darsnack, line 47: I don't think being able to override with a dictionary is as useful when `cardinalities` is still a vector.

```suggestion
function get_emb_sz(cardinalities::Dict{<:Any, <:Integer}, size_overrides=Dict())
    values_and_overrides = map(pairs(cardinalities)) do (col, cardinality)
        cardinality, get(size_overrides, col, nothing)
    end
    get_emb_sz(first.(values_and_overrides), last.(values_and_overrides))
end
```

src/models/tabularmodel.jl
```diff
@@ -0,0 +1,158 @@
...
+    emb_sz_rule(n_cat)
+
+Returns an embedding size corresponding to the number of classes for a 
+categorical variable using the rule of thumb present in python fastai.
+(see https://github.com/fastai/fastai/blob/2742fe844573d06e700f869839fb9ec5f3a9bca9/fastai/tabular/model.py#L12)
+"""
+emb_sz_rule(n_cat) = min(600, round(Int, 1.6 * n_cat^0.56))
+
+"""
+    get_emb_sz(cardinalities, [size_overrides])
```
darsnack, line 11:
```suggestion
    get_emb_sz(cardinalities::AbstractVector, [size_overrides::AbstractVector])
```

src/models/tabularmodel.jl
```diff
@@ -0,0 +1,158 @@
...
+"""
+
+get_emb_sz(cardinalities::AbstractVector{<:Integer}, size_overrides=fill(nothing, length(cardinalities))) =
+    map(zip(cardinalities, size_overrides)) do (cardinality, override)
+        emb_dim = isnothing(override) ? emb_sz_rule(cardinality + 1) : Int64(override)
+        return (cardinality + 1, emb_dim)
+    end
+
+"""
+    get_emb_sz(cardinalities, categorical_cols, [size_overrides])
```
darsnack, line 29:
```suggestion
    get_emb_sz(cardinalities::Dict, [size_overrides::Dict])
```

src/models/tabularmodel.jl
```diff
@@ -0,0 +1,158 @@
...
+
+Returns a collection of tuples containing embedding dimensions corresponding to 
+number of classes in categorical columns present in `cardinalities` and adjusting for NaNs.
```
darsnack, line 32:
```suggestion
Given a map from columns to `cardinalities`, compute the output embedding size according to [`emb_sz_rule`](#).
Return a vector of tuples where each element is `(in_size, out_size)` for an embedding layer.
```

src/models/tabularmodel.jl
```diff
@@ -0,0 +1,158 @@
...
+    get_emb_sz(cardinalities, categorical_cols, [size_overrides])
+
+Returns a collection of tuples containing embedding dimensions corresponding to 
+number of classes in categorical columns present in `cardinalities` and adjusting for NaNs. 
+
+## Keyword arguments
+
+- `size_overrides`: An indexable collection with column name as key and size
+    to override it with as the value.
+- `categorical_cols`: A collection of categorical column names.
```
darsnack, line 38:
```suggestion
- `size_overrides`: An map of output embedding size overrides
                    (i.e. `size_overrides[col]` is the output embedding size for `col`).
```

src/models/tabularmodel.jl
```diff
@@ -0,0 +1,158 @@
...
+"""
+
```
darsnack, line 40:
```suggestion
```

src/models/tabularmodel.jl
```diff
@@ -0,0 +1,158 @@
...
+Returns a collection of tuples containing embedding dimensions corresponding to 
+number of classes in categorical columns present in `cardinalities` and adjusting for NaNs. 
+
+## Keyword arguments
+
+- `size_overrides`: A collection of Integers and `nothing` where the integer present at any index 
+    will be used to override the rule of thumb for getting embedding sizes.
+"""
+
```
darsnack, line 21:
```suggestion
```

src/models/tabularmodel.jl
```diff
@@ -0,0 +1,158 @@
...
+tabular_continuous_backbone(n_cont) = BatchNorm(n_cont)
+
+"""
+    TabularModel(catbackbone, contbackbone, [finalclassifier]; kwargs...)
+
+Create a tabular model which takes in a tuple of categorical values 
+(label or one-hot encoded) and continuous values. The default categorical backbone or `catbackbone` is
+a Parallel of Embedding layers corresponding to each categorical variable, and continuous
+variables are just BatchNormed using `contbackbone`. The output from these backbones is then passed through
+a `finalclassifier` block.
```
darsnack, line 70:
```suggestion
Create a tabular model which operates on a tuple of categorical values 
(label or one-hot encoded) and continuous values.
The categorical backbones (`catbackbone`) and continuous backbone (`contbackbone`) operate on each element of the input tuple.
The output from these backbones is then passed through a series of linear-batch norm-dropout layers before a `finalclassifier` block.
```

src/models/tabularmodel.jl
```diff
@@ -0,0 +1,158 @@
...
+
+## Keyword arguments
+
+- `outsize`: The output size of the final classifier block. For single classification tasks, 
+    this would just be the number of classes and for regression tasks, this could be the
+    number of target continuous variables.
```
darsnack, line 76:
```suggestion
- `outsize`: The output size of the final classifier block. For single classification tasks, 
    this would be the number of classes, and for regression tasks, this would be the
    number of target continuous variables.
```

src/models/tabularmodel.jl
```diff
@@ -0,0 +1,158 @@
...
+- `layersizes`: The sizes of the hidden layers in the classifier block.
```
darsnack, line 77:
```suggestion
- `layersizes`: A vector of sizes for each hidden layer in the sequence of linear layers.
```

src/models/tabularmodel.jl
```diff
@@ -0,0 +1,158 @@
...
+- `dropout_rates`: Dropout probability. This could either be a single number which would be 
+    used for for all the classifier layers, or a collection of numbers which are cycled through
+    for each layer.
```
darsnack, line 80:
```suggestion
- `dropout_rates`: Dropout probabilities for the linear-batch norm-dropout layers.
    This could either be a single number which would be used for for all the layers,
    or a collection of numbers which are cycled through for each layer.
```

src/models/tabularmodel.jl
```diff
@@ -0,0 +1,158 @@
...
+- `batchnorm`: Boolean variable which controls whether to use batch normalization in the classifier.
```
darsnack, line 81:
```suggestion
- `batchnorm`: Set to `false` to skip each batch norm in the linear-batch norm-dropout sequence.
```

src/models/tabularmodel.jl
```diff
@@ -0,0 +1,158 @@
...
+- `activation`: The activation function to use in the classifier layers.
+- `linear_first`: Controls if the linear layer comes before or after BatchNorm and Dropout.
```
darsnack, line 83:
```suggestion
- `linear_first`: Controls if the linear layer comes before or after batch norm and dropout.
```

src/models/tabularmodel.jl
```diff
@@ -0,0 +1,158 @@
...
+        layersizes=(200, 100),
+        kwargs...)
+    TabularModel(catbackbone, contbackbone, Dense(layersizes[end], outsize); layersizes=layersizes, kwargs...)
+end
+
+function TabularModel(
+        catbackbone,
+        contbackbone,
+        finalclassifier;
+        layersizes=[200, 100],
```
darsnack, line 99:
```suggestion
        layersizes=(200, 100),
```

src/models/tabularmodel.jl
```diff
@@ -0,0 +1,158 @@
...
+    number of target continuous variables.
+- `layersizes`: The sizes of the hidden layers in the classifier block.
+- `dropout_rates`: Dropout probability. This could either be a single number which would be 
+    used for for all the classifier layers, or a collection of numbers which are cycled through
+    for each layer.
+- `batchnorm`: Boolean variable which controls whether to use batch normalization in the classifier.
+- `activation`: The activation function to use in the classifier layers.
+- `linear_first`: Controls if the linear layer comes before or after BatchNorm and Dropout.
+"""
+
```
darsnack, line 85:
```suggestion
```

src/models/tabularmodel.jl
```diff
@@ -0,0 +1,158 @@
...
+"""
+    TabularModel(n_cont, outsize, [layersizes; kwargs...])
+
+Create a tabular model which takes in a tuple of categorical values 
+(label or one-hot encoded) and continuous values. The default categorical backbone is
+a Parallel of Embedding layers corresponding to each categorical variable, and continuous
+variables are just BatchNormed. The output from these backbones is then passed through
+a final classifier block. Uses `n_cont` the number of continuous columns, `outsize` which
+is the output size of the final classifier block, and `layersizes` which is a collection of
+classifier layer sizes, to create the model.
```
darsnack, line 137:
```suggestion
Create a tabular model which operates on a tuple of categorical values 
(label or one-hot encoded) and continuous values. The default categorical backbone (`catbackbone`) is
a [`Flux.Parallel`](https://fluxml.ai/Flux.jl/stable/models/layers/#Flux.Parallel) set of `Flux.Embedding` layers corresponding to each categorical variable.
The default continuous backbone (`contbackbone`) is a single [`Flux.BatchNorm`](https://fluxml.ai/Flux.jl/stable/models/layers/#Flux.BatchNorm).
The output from these backbones is concatenated then passed through a series of linear-batch norm-dropout layers before a `finalclassifier` block.

## Arguments

- `n_cont`: The number of continuous columns.
- `outsize`: The output size of the model.
- `layersizes`: A vector of sizes for each hidden layer in the sequence of linear layers.
```

src/models/tabularmodel.jl
```diff
@@ -0,0 +1,158 @@
...
+
+## Keyword arguments
+
+- `cardinalities`: A collection of sizes (number of classes) for each categorical column.
+- `size_overrides`: An optional argument which corresponds to a collection containing 
+    embedding sizes to override the value returned by the "rule of thumb" for a particular index 
+    corresponding to `cardinalities`, or `nothing`.
+"""
+
```
darsnack, line 146:
```suggestion
```

darsnack (changes requested):
Mostly docstring changes.

darsnack (approved):

darsnack:
@lorenzoh this is good to go. I'll let you merge to have the final skim.

PR #141, opened 2021-07-16
Add tabular learning methods

manikyabard:
Adds the implementation of DLPipelines.jl `LearningMethod` interface for tabular regression and tabular classification tasks.


src/methods/tabularclassification.jl
```diff
@@ -0,0 +1,79 @@
...
+
+function DLPipelines.encode(method::TabularClassification, context, input)
+    tempinp = (; zip(method.columns, collect(input))...)
+    item = DataAugmentation.TabularItem(tempinp, method.columns)
+    tfminp = run(method.tfms, context, item)
+    x = (
+            [Int32(tfminp.data[col]) for col in method.catcols], 
+            [tfminp.data[col] for col in method.contcols]
+    )
+    y = Flux.onehot(tfminp.data[method.targetcol], method.targetclasses)
```
lorenzoh, line 33: Move this into `TabularTransforms` so it can be reused

lorenzoh:
_Some notes on integration with the data block API_

Ideally the table-specific data processing functionality would be encapsulated in an `Encoding` called `TabularPreprocessing` that works on a `TableRow` `Block`. If the target variable is separated from the row, we could then write the learning tasks as:

```julia
# tabular classification
BlockMethod(
    (TableRow(catcols, contcols), Label(classes)),
    (
        TabularPreprocessing(),  # only transforms `TableRow`
        OneHot(),  # only transforms `Label`
    )
)

# tabular regression
BlockMethod(
    (TableRow(catcols, contcols), Continuous(n)),
    (
        TabularPreprocessing(),  # only transforms `TableRow`
    )
)
```

src/datablock/block.jl
```diff
@@ -155,3 +155,47 @@ function checkblock(
...
+end
+
+function TableRow(catcols, contcols, categorydict)
+    TableRow{length(catcols), length(contcols)}(catcols, contcols, categorydict)
+end
+
+function checkblock(block::TableRow, x)
+    (all(col -> x[col] ∈ block.categorydict[col], block.catcols) &&
+    all(col -> x[col] isa Number, block.contcols) &&
+    all(col -> haskey(block.categorydict, col), block.catcols))
```
lorenzoh, line 183: This will error when a column is missing.

src/datablock/block.jl
```diff
@@ -155,3 +155,47 @@ function checkblock(
...
+
+"""
+    Continuous(n) <: Block
+
+`Block` for collections of numbers. `data` is valid if it's 
+length is `n` and contains `Number`s.
+"""
+
+struct Continuous <: Block
+    n
```
lorenzoh, line 196: This should be concretely typed:
```suggestion
    size::Int
```

src/datablock/block.jl
manikyabard: Should be fixed now.

src/datablock/block.jl
```diff
@@ -155,3 +155,49 @@ function checkblock(
...
+    TableRow{M, N}(catcols, contcols, categorydict) <: Block
+
+`Block` for table rows with M categorical and N continuous columns. `data`
+is valid if it satisfies the `AbstractRow` interface in Tables.jl, values 
+present in indices for categorical and continuous columns are consistent, 
+and `data` is indexable by the elements of `catcols` and `contcols`.
+"""
+struct TableRow{M, N} <: Block
+    catcols
+    contcols
```
lorenzoh, line 172:
```suggestion
    catcols::NTuple{M}
    contcols::NTuple{N}
```

src/datablock/encoding.jl
```diff
@@ -246,7 +246,7 @@ function testencoding(encoding, block, data = mockblock(block))
         @test checkblock(outblock, outdata)
 
         # Test decoding (if supported) works correctly
-        inblock = decodedblock(encoding, outblock, true)
+        inblock = decodedblock(encoding, outblock)
```
lorenzoh, line 249: Why is this changed?

src/datablock/models.jl
```diff
@@ -59,3 +59,38 @@ function blockmodel(inblock::ImageTensor{N}, outblock::KeypointTensor{N}, backbo
     head = Models.visionhead(outch, prod(outblock.sz)*N, p = 0.)
     return Chain(backbone, head)
 end
+
+"""
+    blockmodel(inblock::TableRow{M, N}, outblock::Union{Continuous, OneHotTensor{0}}, backbone=nothing) where {M, N}
```
lorenzoh, line 64: Can we make it clear that `backbone` is an optional argument and what the default is?

src/encodings/tabularpreprocessing.jl
```diff
@@ -0,0 +1,117 @@
+"""
+    EncodedTableRow{M, N} <: Block
+
+Block for processed rows having a tuple of M categorical and 
+N continuous value collections.
+"""
+
+struct EncodedTableRow{M, N} <: Block
+    catcols
+    contcols
```
lorenzoh, line 10:
```suggestion
    catcols::NTuple{M}
    contcols::NTuple{N}
```

src/encodings/tabularpreprocessing.jl
```diff
@@ -0,0 +1,117 @@
...
+    categorydict
+end
+
+function EncodedTableRow(catcols, contcols, categorydict)
+    EncodedTableRow{length(catcols), length(contcols)}(catcols, contcols, categorydict)
+end
+
+function checkblock(::EncodedTableRow{M, N}, x) where {M, N}
```
lorenzoh, line 18: Maybe we should restrain `x` to be a 2-tuple. Unless there is a reason not to?

src/encodings/tabularpreprocessing.jl
```diff
@@ -0,0 +1,117 @@
...
+Encodes `TableRow`s by applying a composition of various preprocessing
+steps available in `DataAugmentation.jl`. Currently, preprocessing could 
+consist of
+- [`DataAugmentation.NormalizeRow`](#) (for normalizing a row of data for continuous columns)
+- [`DataAugmentation.FillMissing`](#) (for filling missing values)
+- [`DataAugmentation.Categorify`](#) (for label encoding categorical columns, which can be later used for indexing into embedding matrices)
+or a sequence of these transformations.
+
+"""
+
```
lorenzoh, line 34:
```suggestion
```

src/encodings/tabularpreprocessing.jl
```diff
@@ -0,0 +1,117 @@
...
+function checkblock(::EncodedTableRow{M, N}, x) where {M, N}
+    length(x[1]) == M && length(x[2]) == N
+end
+
+"""
+    TabularTransform <: Encoding
+
+Encodes `TableRow`s by applying a composition of various preprocessing
+steps available in `DataAugmentation.jl`. Currently, preprocessing could 
+consist of
```
lorenzoh, line 27:
```suggestion
Encodes a `TableRow` by applying the following preprocessing steps:
```

src/encodings/tabularpreprocessing.jl
```diff
@@ -0,0 +1,117 @@
...
+    (catvals, contvals)
+end
+
+"""
+The helper functions defined below can be used for quickly constructing a dictionary,
+which will be required for creating various tabular transformations available in DataAugmentation.jl.
+
+These functions assume that the table in the TableDataset object td has Tables.jl columnaccess interface defined.
+"""
+
```
lorenzoh, line 64:
```suggestion
```

src/encodings/tabularpreprocessing.jl
```diff
@@ -0,0 +1,117 @@
...
+function getcoltypes(td::Datasets.TableDataset)
+    schema = Tables.schema(td.table)
+    contcols = filter(col->schema.types[findfirst(isequal(col), schema.names)] <: Union{Number, Missing}, schema.names)
+    catcols = filter(col->!(col in contcols), schema.names)
+    catcols, contcols
+end
+
+function gettransforms(td::Datasets.TableDataset)
```
lorenzoh, line 105: Can we use this in a constructor for `TabularTransform` so you can call it like `TabularTransform(TableDataset(...))`?

src/encodings/tabularpreprocessing.jl
```diff
@@ -0,0 +1,117 @@
...
+    catcols, contcols = getcoltypes(td)
+    normstats = FastAI.gettransformationdict(td, DataAugmentation.NormalizeRow, contcols)
+    fmvals = FastAI.gettransformationdict(td, DataAugmentation.FillMissing, contcols)
+    catdict = FastAI.gettransformationdict(td, DataAugmentation.Categorify, catcols)
+    
+    normalize = DataAugmentation.NormalizeRow(normstats, contcols)
+    categorify = DataAugmentation.Categorify(catdict, catcols)
+    fm = DataAugmentation.FillMissing(fmvals, contcols)
+    
+    fm|>normalize|>categorify
```
lorenzoh, line 115:
```suggestion
    return fm |> normalize |> categorify
```

src/encodings/onehot.jl
```diff
@@ -24,6 +24,7 @@ function checkblock(block::OneHotTensorMulti{N}, a::AbstractArray{T, M}) where {
     return N + 1 == M && last(size(a)) == length(block.classes)
 end
 
+
```
lorenzoh, line 27:
```suggestion
```

lorenzoh (changes requested):

src/datablock/block.jl
```diff
@@ -155,3 +155,49 @@ function checkblock(
...
+    categorydict
+end
+
+function TableRow(catcols, contcols, categorydict)
+    TableRow{length(catcols), length(contcols)}(catcols, contcols, categorydict)
+end
+
+function checkblock(block::TableRow, x)
+    columns = Tables.columnnames(x)
+    (all(col -> col ∈ columns, [block.catcols..., block.contcols...]) &&
```
darsnack, line 182:
```suggestion
    (all(col -> col ∈ columns, (block.catcols..., block.contcols...)) &&
```

src/datablock/models.jl
```diff
@@ -59,3 +59,38 @@ function blockmodel(inblock::ImageTensor{N}, outblock::KeypointTensor{N}, backbo
...
+
+Contruct a model for tabular classification or regression. `backbone` should be a 
+Dict of categoricalbackbone, continuousbackbone, and a finalclassifier layer, with 
+the first two taking in batches of corresponding row value matrices.
+"""
+
+function blockmodel(
+        inblock::EncodedTableRow{M, N}, 
+        outblock::Union{Continuous, OneHotTensor{0}}, 
+        backbone=Dict()) where {M, N}
```
darsnack, line 74: Prefer to use a `NamedTuple` over a `Dict`.

src/datablock/models.jl
```diff
@@ -59,3 +59,38 @@ function blockmodel(inblock::ImageTensor{N}, outblock::KeypointTensor{N}, backbo
...
+    end
+    if !(:continuousbackbone in backbonekeys)
+        backbone[:continuousbackbone] = Models.tabular_continuous_backbone(N)
+    end
+    if (:finalclassifier in backbonekeys)
+        TabularModel(backbone[:categoricalbackbone], backbone[:continuousbackbone], backbone[:finalclassifier])
+    else
+        outsz = outblock isa Continuous ? outblock.size : length(outblock.classes)
+        TabularModel(backbone[:categoricalbackbone], backbone[:continuousbackbone]; outsz=outsz)
+    end
```
darsnack, line 92: Could simplify this:
```julia
    default￼￼￼_backbones = (categorical = ...,
                         continuous = ...,
                         classifier = ...)
    backbones = [haskey(backbone, k) ? backbone[k] : default_backbones[k]
                 for k in (:categorical, :continuous, :classifier)]
    TabularModel(backbones...)
```

src/encodings/tabularpreprocessing.jl
```diff
@@ -0,0 +1,117 @@
...
+end
+
+"""
+The helper functions defined below can be used for quickly constructing a dictionary,
+which will be required for creating various tabular transformations available in DataAugmentation.jl.
+
+These functions assume that the table in the TableDataset object td has Tables.jl columnaccess interface defined.
+"""
+
+function gettransformationdict(td, ::Type{DataAugmentation.NormalizeRow}, cols)
```
darsnack, line 65: `gettransformdict` to be shorter?

darsnack (changes requested):

src/datablock/encoding.jl
manikyabard: For the encodings for which decoding isn't supported yet, like `TabularTransforms` (I'll rename this to TabularPreprocessing), it will give incorrect results I think. 
The encoded block after applying `TabularTransforms` on a `TableRow` is `EncodedTableRow`, and if decoding was supported, then the decoded block should be a `TableRow` again. But as it isn't supported, using the fill option would make it so that `EncodedTableRow` is returned again (instead of `nothing` through the fallback method) which wouldn't be correct. 
The tests after this could fail because `inblock` (which would be the encoded block in this situation), doesn't have to be the same as the original block. Also I'm not sure if the `if` condition would ever be `false` as of now.

src/datablock/models.jl
manikyabard: Sure I'll do that for `blockmodel`. I was wondering if there is way for `backbone` to still be optional if someone is using `methodmodel` or maybe even `methodlearner`.

src/encodings/tabularpreprocessing.jl
manikyabard: Yeah, I think we can do that.

src/encodings/tabularpreprocessing.jl
manikyabard: Sure, so something like this?
```julia
TabularTransform(td::TableDataset) = TabularTransform(gettransforms(td))
```

lorenzoh (commented):
Changes look good 👍

lorenzoh:
Anything else that is missing before this can be merged?

Only thing I can think of are simple `BlockMethod` wrappers like those in [`src/fasterai/learningmethods.jl`](https://github.com/FluxML/FastAI.jl/blob/master/src/fasterai/learningmethods.jl)? Each should have a constructor that takes just the `blocks` i.e. `::Tuple{<:TableRow, <:Continuous}` and one convenience constructor. They can then be registered with some block types so you can find them using `findlearningmethods`(s. the same file for how).

src/datablock/block.jl
```diff
@@ -155,3 +155,49 @@ function checkblock(
...
+`Block` for table rows with M categorical and N continuous columns. `data`
+is valid if it satisfies the `AbstractRow` interface in Tables.jl, values 
+present in indices for categorical and continuous columns are consistent, 
+and `data` is indexable by the elements of `catcols` and `contcols`.
+"""
+struct TableRow{M, N} <: Block
+    catcols::NTuple{M}
+    contcols::NTuple{N}
+    categorydict
+end
```
darsnack, line 174: Not sure if FastAI.jl prefers to not use concrete fields @lorenzoh 

```suggestion
struct TableRow{M, N, T} <: Block
    catcols::NTuple{M}
    contcols::NTuple{N}
    categorydict::T
end
```

src/encodings/tabularpreprocessing.jl
```diff
@@ -0,0 +1,113 @@
...
+        dict[col] = unique(vals)
+    end
+    dict
+end
+
+"""
+The convenience functions defined below can be used to get the categorical and continuous 
+columns present in a `TableDataset`, and quickly get transformations for it which can be 
+passed in `TableTransform`.
+"""
```
darsnack, line 92: Turn this into docstrings (preferred) or a comment?

src/encodings/tabularpreprocessing.jl
```diff
@@ -0,0 +1,113 @@
...
+Encodes a `TableRow` by applying the following preprocessing steps:
+- [`DataAugmentation.NormalizeRow`](#) (for normalizing a row of data for continuous columns)
+- [`DataAugmentation.FillMissing`](#) (for filling missing values)
+- [`DataAugmentation.Categorify`](#) (for label encoding categorical columns, which can be later used for indexing into embedding matrices)
+or a sequence of these transformations.
+
+"""
+struct TabularTransform <: Encoding
+	tfms
+end
```
darsnack, line 34: Same as before.

```suggestion
struct TabularTransform{T} <: Encoding
	tfms::T
end
```

src/encodings/tabularpreprocessing.jl
```diff
@@ -0,0 +1,113 @@
...
+
+Block for processed rows having a tuple of M categorical and 
+N continuous value collections.
+"""
+
+struct EncodedTableRow{M, N} <: Block
+    catcols::NTuple{M}
+    contcols::NTuple{N}
+    categorydict
+end
```
darsnack, line 12: Same comment about concrete field types

src/datablock/models.jl
darsnack: Could even have a helper `default_tabular_backbones()` that makes the default `NamedTuple` in [this suggestion](https://github.com/FluxML/FastAI.jl/pull/141/files#r689111096)

darsnack (changes requested):
Some more small nitpicks

manikyabard:
I think the tests are failing because the model PR isn't merged yet.

darsnack:
Is there a reason that `decode(::Encoding, ::Block)` defaults to `nothing` (same for `encode`)? I get that it allows us to distinguish between when a block is modified but happens to be the same type vs. when it is not affected by `Encoding`. But is there value in tracking that information? Could we just default to an identity map like
```julia
encodedblock(::Encoding, block::Block) = block
decodedblock(::Encoding, block::Block) = block
```
This seems intuitive to me. Unless we use the `nothing` information later in the pipeline.

lorenzoh:
> Is there a reason that `decode(::Encoding, ::Block)` defaults to `nothing` (same for `encode`)? I get that it allows us to distinguish between when a block is modified but happens to be the same type vs. when it is not affected by `Encoding`. But is there value in tracking that information? Could we just default to an identity map like
> 
> ```julia
> encodedblock(::Encoding, block::Block) = block
> decodedblock(::Encoding, block::Block) = block
> ```
> 
> This seems intuitive to me. Unless we use the `nothing` information later in the pipeline.

That's the only way to know that data was not changed without running the encoding. `encode` and `decode` themselves do have an identity as the default, `encodedblock` and `decodedblock` just handle the block info.

lorenzoh:
This looks good.

Are the tests failing due to the issue with the decoding not being implemented? In that case, we could add a definition for just`decodedblock` without implementing `decode`. Or add a `decode = true` kwarg to `testencoding` so that it only does the decoding tests if `decode == true`.

manikyabard:
> This looks good.
> 
> Are the tests failing due to the issue with the decoding not being implemented? In that case, we could add a definition for just`decodedblock` without implementing `decode`. Or add a `decode = true` kwarg to `testencoding` so that it only does the decoding tests if `decode == true`.

I think the tests might be failing because `methodmodel` is being tested as well and the model code isn't present in this branch yet.

lorenzoh (approved):

lorenzoh:
maybe some minor doc improvements could still be made, but let's do that in another PR that also integrates the notebook with the documentation. We still have time for that before the next release.

PR #149, opened 2021-08-09
Fix Pkg usage in developing.md

amqdn:
Hi!

First PR. Happy to file an issue first if necessary. 

In following the instructions in `developing.md`, I found that Pkg will not accept a single URL arg in some of its methods (FYI, this goes for a single path arg as well). I've made all the changes to make the instructions work as I followed them. 

Open to comments and edits!

### DIFF
--- a/developing.md
+++ b/developing.md
@@ -10,7 +10,7 @@ You'll have to do the following:
 
 ```julia
 using Pkg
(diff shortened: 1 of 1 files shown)


lorenzoh:
Thanks for the fix!

PR #167, opened 2021-09-03
update get_emb_sz method

manikyabard:
Updates `get_emb_sz` to ensure the order of embedding sizes is as expected which might not have been the case with just using `Dict` key order.

### DIFF
--- a/notebooks/tabularclassification.ipynb
+++ b/notebooks/tabularclassification.ipynb
@@ -18,7 +18,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 2,
+   "execution_count": 1,
    "id": "bc1f44bd",
    "metadata": {},
    "outputs": [],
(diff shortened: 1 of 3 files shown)


darsnack:
Code changes look fine, but I am wondering whether we should just drop the `Dict` version. I get that it is easier to specify a single column override with the `Dict`, but it is really not that hard to generate a vector of overrides + `nothing` with a generator. It's so deep into the API hierarchy too that we can forgo the same level of convenience we apply to the top level functions.

And if the consuming model building function always expects an in-order vector, then the `Dict` approach doesn't really make sense.

manikyabard:
Yeah, that does make sense. I think I'll then just remove the method which takes in a `Dict`, and update the tabularclassification notebook to show an example of using custom embedding sizes for people new to FastAI.jl or julia.

lorenzoh:
Thanks Manikya for the change! Is this ready to merge?

manikyabard:
If everything looks okay with the changes, then yeah it should be.

darsnack (approved):

PR #171, opened 2021-09-16
Setup step for encodings

lorenzoh:
This PR adds functionality for setting up encodings that need access to a data container, e.g. to compute some statistics over each observation.

The interface looks something like this:

```julia
"""
    setup(Encoding, block, data; kwargs...)

Create an encoding using statistics derived from a data container `data`
with observations of block `block`. Used when some arguments of the encoding
are dependent on the dataset. `data` should be the training dataset. Additional
`kwargs` are passed through to the regular constructor of `Encoding`.

## Examples

``julia
(images, labels), blocks = loaddataset("imagenette2-160", (Image, Label))
setup(ImagePreprocessing, Image{2}(), images; buffered = false)
``

``julia
data, block = loaddataset("adult_sample", TableRow)
setup(TabularPreprocessing, block, data)
``
"""
function setup end
```

Beside the `setup` implementation, this addition will also change the interface of the learning method functions like `ImageClassificationSingle` as they now take a data container `data` as a second argument in case the setup step is needed.

Specifically, the computer vision methods can compute normalization stats on the image (which is turned off by default for performance reasons; there'll be a flag) and the tabular methods compute stats for `TabularPreprocessing` on a table dataset.

To-Dos:

- implement, test, and document `setup` for
  - [x] `ImagePreprocessing`
  - [x] `TabularPrepreprocessing`
- [x] update learning method functions to use setup
- docs
  - [x] update quickstart docs to include tabular tasks (this was pending on this functionality)
  - [x] update tabular clf notebook
- [x] update CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased] - 2020-07-32 – now
 
 ### Added
+
 - High-level API "FasterAI"
+    - dataset recipes
+    - learning method helpers
     - Find datasets and learning methods based on `Block`s: `finddataset`, `findlearningmethods`
     - `loaddataset` for quickly loading data containers from configured recipes
 - Data container recipes (`DatasetRecipe`, `loadrecipe`)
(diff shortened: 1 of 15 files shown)


lorenzoh:
Added code for image statistics setup, on `"imagenette2-160"` takes around 2 secs using threads to go through the whole dataset.

PR #175, opened 2021-09-24
change julia compat to 1.6

EvoArt:
Data loading operations involve taking views of Strings. This errors with pre-1.6 Julia.


lorenzoh:
Don't have the incompatible code right now, can you point me to the location? Maybe there's an easy fix to remain backward compatible. Otherwise I'll merge this 👍

lorenzoh:
Seeing as 1.6 is the new LTS now, I'll just go ahead and merge this.

Issue #165, opened 2021-08-31
Support for non-supervised learning tasks


lorenzoh:
So right now, much of the data block magic is specialized on supervised learning tasks, e.g. `blocklossfn`, `blockmodel` and the visualizations. It would be nice to better support other kinds of tasks as well. The underlying APIs are already pretty general, e.g. `encode` has no restraints on number of blocks encodings. It's more that right now, `BlockMethod` assumes a tuple `(inputblock, targetblock)` and uses that to determine what block the model output is and what blocks are in a batch for visualization.

For the FastAI Q&A @darsnack prepared a [variational autoencoder tutorial](https://github.com/darsnack/fastai-vae-demo) where the data is really just one image that itself is the target and we ended up having to overwrite some methods to get it to work.

So, the thing that is mostly different for different learning tasks, is how each block of data is used. Below I try to summarize the different kinds of blocks that need to be known to use the block method APIs.

- `sampleblock` i.e. what are the blocks of one sample
- `outputblock` i.e. what block is output by the model
- `xblock` i.e. what block is input to the model
- `yblock` i.e. what block does the loss function compare `outputblock` to
- `showpredblocks` i.e. what blocks are compared when using `showpredictions`
- `showoutputblocks` i.e. what blocks are compared when using `showoutputs`

For a basic supervised learning method, we have

- `sampleblock = (inputblock, targetblock) = method.blocks`
- `outputblock = yblock` (can be overwritten)
- `xblock, yblock = encode(sampleblock)`
- `predblock = decode(yblock)` (usually same as `targetblock`)
- `showoutputblocks = (xblock, yblock, outputblock)`
- `showpredblocks = (inputblock, targetblock, predblock) = decode(showoutputblocks)`

For the autoencoder example, we have

- `sampleblock = inputblock = targetblock = predblock`
- `outputblock = xblock = yblock = encode(sampleblock)`
- `showpredblocks = (inputblock, predblock)`
- `showoutputblocks = (xblock, outputblock)`

This is how the blocks are used:

- core interface
    - `encode() = encode(sampleblock)`
    - `encodeinput() = encode(inputblock)`
    - `encodetarget() = encode(targetblock)`
    - `decodeypred() = decode(outputblock)`
    - `decodex() = decode(xblock)`
    - `decodey() = decode(yblock)`
- training interface
    - `methodlossfn() = blocklossfn(outputblock, yblock)`
[...]

PR #188, opened 2021-12-20
Make block learning methods more modular

lorenzoh:
Progress for tackling #165. 

[VAE notebook](https://github.com/FluxML/FastAI.jl/blob/lorenzoh/modular-methods/notebooks/vae.ipynb) from @darsnack https://github.com/darsnack/fastai-vae-demo.


darsnack:
I took a look over the updated notebook. The FastAI.jl hooks make sense to me, and it seems like the boilerplate has been reduced quite a bit from the original example. The only thing that's confusing for me is why `EmbeddingMethod` requires so many blocks (e.g. `x`, `y`, `ŷ`, `sample`, etc.). Maybe an explanation or comment in the code describing their purpose would help?

In terms of the tutorial aspect, I would add something at the start giving an overview of each piece of FastAI.jl that we will override/extend. Something like:
```
The FastAI.jl default methods and training loops are designed for supervised learning. To support our unsupervised approach, we will extend FastAI.jl by
- creating a data iterator over only the samples (no labels)
- defining our own block method, `EmbeddingMethod` (the analogue to `SupervisedMethod`)
- defining our own FluxTraining.jl phase, `VAETrainingPhase`, which will take the gradients with respect to our unsupervised loss correctly
```

lorenzoh:
Thanks for the feedback! Planning to either (1) put the source for `EmbeddingMethod` into FastAI.jl itself or (2) make the definition shorter, using just a function with only the blocks needed for this tutorial, e.g.

```julia
function EmbeddingMethod(block, encodings)
    sample = block
    x = y\hat = encodedsample = FastAI.encodedblockfilled(encodings, block)
    return BlockMethod((; sample, x, y\hat, encodedsample), encodings)
end
```

With then some explanation what those blocks mean and maybe where they're needed (i.e. `showoutputs` needs `y\hat`)

darsnack:
> With then some explanation what those blocks mean and maybe where they're needed (i.e. showoutputs needs y\hat). Would that make the part clearer?

Yes, that makes it much clearer!

PR #193, opened 2022-01-24
Add `imdb_sample` recipe

Chandu-4444:
Hey, I tried to add the `imdb_sample` recipe to Tabular Datasets. Also, I added the correct hash value for the dataset to download.

Umm... Are there any plans to make a `/text` folder (like fastai has) to put all the code related to text?


lorenzoh:
Looks good to me!

> Umm... Are there any plans to make a /text folder (like fastai has) to put all the code related to text?

So I recently factored out all the vision- and tabular- related functionality into submodules `Vision` and `Tabular`. FastAI.jl doesn't yet have support for NLP tasks (though it could be integrated with the JuliaText ecosystem and Transformers.jl) so there hasn't been the need for a specific submodule yet.

Issue #153, opened 2021-08-10
Dataset recipes


lorenzoh:
With #151, FastAI.jl is getting high-level interfaces for searching datasets (`finddatasets`) and loading datasets into task-specific data containers (`loaddataset`). There is also a new `DatasetRecipe` that encapsulates configuration for loading a data container and the block information from a path. These recipes can be registered with a dataset so that they can be found using the above high-level functions.

The fastai dataset colletion comes with quite a lot of datasets, so only a few have recipes yet. This issue tracks the progress on adding recipes to all the datasets. Contributions of recipe types and recipe configs for datasets are welcome. See `src/datasets/recipes.jl` for example recipe implementations and `src/datasets/fastairegistry` for how recipes are registered.

## Progress

- [ ] CUB_200_2011
- [ ] bedroom
- [ ] caltech_101
- [ ] cifar10
- [ ] cifar100
- [ ] food-101
- [ ] imagenette-160
- [ ] imagenette-320
- [ ] imagenette
- [ ] imagenette2-160
- [ ] imagenette2-320
- [ ] imagenette2
- [ ] imagewang-160
- [ ] imagewang-320
- [ ] imagewang
- [ ] imagewoof-160
- [ ] imagewoof-320
- [ ] imagewoof
- [ ] imagewoof2-160
- [ ] imagewoof2-320
- [ ] imagewoof2
- [ ] mnist_png
- [ ] mnist_var_size_tiny
- [ ] oxford-102-flowers
- [ ] oxford-iiit-pet
- [ ] stanford-cars
- [ ] ag_news_csv
- [ ] amazon_review_full_csv
- [ ] amazon_review_polarity_csv
- [ ] dbpedia_csv
- [ ] giga-fren
- [ ] imdb
- [ ] sogou_news_csv
- [ ] wikitext-103
- [ ] wikitext-2
- [ ] yahoo_answers_csv
- [ ] yelp_review_full_csv
- [ ] yelp_review_polarity_csv
- [ ] biwi_head_pose
- [ ] camvid
- [ ] pascal-voc
- [ ] pascal_2007
- [ ] pascal_2012
- [ ] siim_small
- [ ] skin-lesion
- [ ] tcga-small
- [ ] adult_sample
- [ ] biwi_sample
- [ ] camvid_tiny
- [ ] dogscats
- [ ] human_numbers
- [ ] imdb_sample
- [ ] mnist_sample
- [ ] mnist_tiny
[...]

PR #194, opened 2022-01-25
Add food-101 recipe

Chandu-4444:
I've added `food-101` to the RECIPES

Ref: #153


lorenzoh:
Have you tested loading this into a data container and can you check that `checkblock(getobs(data, 1), blocks)` is `true`?

Chandu-4444:
Oh, I forgot about tests. I'll check and will update the tests.

lorenzoh:
You don't need to write test cases for this, just test it manually. Unfortunately, we can't test every dataset recipe all the time since then CI would have to download every dataset (and just this one is a few GB I think)

Chandu-4444:
Okay, I'll test it manually and will update it here then 👍

Chandu-4444:
> Have you tested loading this into a data container and can you check that `checkblock(getobs(data, 1), blocks)` is `true`?

Yeah, it returned true.

lorenzoh:
Thanks for checking!

Issue #155, opened 2021-08-11
Timeseries regression and classification usecase


AriMKatz:
Hi,

See https://github.com/timeseriesAI/tsai for a TS package into of FastAI python.  

Just adding this to the wishlist. I might have some time to work on it (hopefully) for an upcoming project  (or maybe will have to use the python package depending on cycles). 

The salient question now is, are the mid-level abstractions sufficiently generic  for the TS usecase and if not what needs to change to support that?

I'll try to dig into that package this week to see what sort of functionality it needs.

lorenzoh:
I think the mid-level abstractions are sufficiently generic 👍 

I'm not familiar with the SOTA in time series, but here's a list of things that would need to be implemented for supervised time-series learning tasks:

- a `TimeSeries` `Block` type. Used for visualization creating models automatically, data validation, `BlockMethod` creation, dataset discovery and more... 
- time-series datasets that can be downloaded using `datasetpath`
- Dataset recipes that load those datasets into data containers usable for learning
- if needed, `Encoding`s that transform/augment the time series data

Would be great to have time series as an application in FastAI.jl, so let me know if you'd like more info on some of these points. See [`block.jl`](https://github.com/FluxML/FastAI.jl/blob/master/src/datablock/block.jl) and [`encoding.jl`](https://github.com/FluxML/FastAI.jl/blob/master/src/datablock/encoding.jl) for a bit more info on what implementing a `Block` entails.

ToucheSir:
Last I checked, TSAI integration with fastai proper was kind of tenuous, so we could definitely do better here. My big caveat is that time series is just as wide of a domain as images/computer vision, but less explored and nit nearly as well documented. i.e. something that would take at least a GSoC project's worth of work to get a satisfactory start on.

codeboy5:
I was working on creating a time series block. This covers some of functionality we require. How does this look ?

```
struct TimeSeries{M,N} <: Block end

function checkblock(::TimeSeries{M,N}, obs::AbstractArray{T,2}) where {M,N,T<:Number}
    size(obs) == (M,N)
end

mockblock(::TimeSeries{M,N}) where {M,N} = rand(Float64, (M,N))    

function setup(::Type{TimeSeries}, data)
    # N,M = size(data[1,:,:])
    N,M = size(getobs(data, 1))
    return TimeSeries{N,M}()
end

function showblock!(io, ::ShowText, block::TimeSeries, obs)
[...]

PR #199, opened 2022-03-04
Added Time Series Container and Block

codeboy5:
Added Time Series Container and Block. Currently can only load univariate time series. This is work in progress for issue #155 .
I was planning to add `loaddataset` function for such datasets. Currently all datasets have the same root URL :- "https://s3.amazonaws.com/fast-ai-" . For time series datasets the root url is different so i think we can proceed by add `root_url` field in the `FastAIDataset` structure.
How does this sound ?

### DIFF
--- a/src/TimeSeries/blocks/timeseries.jl
+++ b/src/TimeSeries/blocks/timeseries.jl
@@ -1,0 +1,24 @@
+
+struct TimeSeries{M,N} <: Block end
+
+function checkblock(::TimeSeries{M,N}, obs::AbstractArray{T,2}) where {M,N,T<:Number}
+    size(obs) == (M,N)
+end
+
+mockblock(::TimeSeries{M,N}) where {M,N} = rand(Float64, (M,N))    
+
+function setup(::Type{TimeSeries}, data)
+    # N,M = size(data[1,:,:])
+    N,M = size(getobs(data, 1))
+    return TimeSeries{N,M}()
+end
+
+# visualization
+
+function showblock!(io, ::ShowText, block::TimeSeries, obs)
+    plot = UnicodePlots.lineplot(obs[1,:])
+    for j=2:size(obs,1)
+        UnicodePlots.lineplot!(plot, obs[j,:])
+    end
+    print(io, plot)
+end
--- a/src/datasets/Datasets.jl
+++ b/src/datasets/Datasets.jl
@@ -63,6 +63,7 @@ export
     # primitive containers
     FileDataset,
     TableDataset,
+    TimeSeriesDataset,
 
     # utilities
     isimagefile,
--- a/src/datasets/containers.jl
+++ b/src/datasets/containers.jl
@@ -91,6 +91,26 @@ LearnBase.nobs(dataset::TableDataset{<:DataFrame}) = nrow(dataset.table)
 LearnBase.getobs(dataset::TableDataset{<:CSV.File}, idx) = dataset.table[idx]
 LearnBase.nobs(dataset::TableDataset{<:CSV.File}) = length(dataset.table)
 
+# TimeSeries Dataset
+
+struct TimeSeriesDataset{}
+    table::AbstractArray{Float64,3}
+end 
+
+function TimeSeriesDataset(path::AbstractPath)
+    mat = Matrix(DataFrame(CSV.File(path,header=0)))
+    N,M = size(mat)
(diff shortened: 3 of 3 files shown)


codeboy5:
I am also changing so that we can read the data from `.arff` files instead of `.tsv` files since most datasets are available in that format.

codeboy5:
This allows us to run the following code, currently only uni variate time series are supported. The block and container will work for multi variate too but the process to read the input would change somewhat.

```julia
using FastAI

path = "/home/saksham/Desktop/GSOC/Adiac"
recipe = TimeSeries.TimeSeriesDatasetRecipe(file="Adiac_TRAIN.arff")
data, blocks = loadrecipe(recipe, path)
println(Datasets.testrecipe(recipe, data, blocks))
sample = getobs(data,5)
println(checkblock(blocks, sample))
```

which returns

```
Test Passed
true
```

codeboy5:
This allows us to do the following
```julia
path = datasetpath("adiac")
recipe = TimeSeries.TimeSeriesDatasetRecipe(file="Adiac_TRAIN.arff")
data, blocks = loadrecipe(recipe, path)
```

src/datasets/fastaidatasets.jl
```diff
@@ -105,6 +107,9 @@ const DATASETCONFIGS = [
     FastAIDataset("annotations_trainval2017", "coco", datadepname="coco-annotations_trainval2017", extension="zip"),
     FastAIDataset("stuff_annotations_trainval2017", "coco", datadepname="coco-stuff_annotations_trainval2017", extension="zip"),
     FastAIDataset("panoptic_annotations_trainval2017", "coco", datadepname="coco-panoptic_annotations_trainval2017", extension="zip"),
+
+    # time series
+    FastAIDataset("Adiac", "", "9c808bcfc77f3cab0a640bfa23e680ccde91da6981962e63c0f4f0737dcf6390", datadepname="adiac", extension="zip", root_url = "http://www.timeseriesclassification.com/Downloads", subpath="")
```
lorenzoh, line 112: Size missing here if available

lorenzoh (commented):
Looking very good so far! I've left some comments. Unfortunately I am not familiar with time series using deep learning, so I don't know what the observations should look like and what encodings will be needed. Do you have a reference that gives an overview of these things?

Other than that, some tests would be great :) (see the rest of the library as reference for how to write these inline, and check [ReTest.jl](https://github.com/JuliaTesting/ReTest.jl) for how to run these interactively); dataset doesn't need a test since we don't want to download that on every CI run)

codeboy5:
> Looking very good so far! I've left some comments. Unfortunately I am not familiar with time series using deep learning, so I don't know what the observations should look like and what encodings will be needed. Do you have a reference that gives an overview of these things?
> 
> Other than that, some tests would be great :) (see the rest of the library as reference for how to write these inline, and check [ReTest.jl](https://github.com/JuliaTesting/ReTest.jl) for how to run these interactively); dataset doesn't need a test since we don't want to download that on every CI run)

Let me see if I can find a good reference for this. This has some (https://timeseriesai.github.io/tsai/data.preparation.html) which we can refer to for encoding. 
I'll work on adding tests and look at resolving some of the comments you mentioned. I think going further we would also need to decide some models that we want to work on first.

lorenzoh:
That's a good resource and probably a good reference for implementations here :+1: 

The best way to proceed is to take an example task from tsai, e.g. classification, and implement the components (incl. the model) used there.

codeboy5:
Hey sorry, I was busy with some school stuff. 
So i am currently working on [this tutorial](https://github.com/timeseriesAI/tsai/blob/main/tutorial_nbs/01_Intro_to_Time_Series_Classification.ipynb)  and add this functionality to FastAI.jl .

lorenzoh:
That seems like a great starting point!

> Hey sorry, I was busy with some school stuff.

No worries, it's normal that PRs take time and are punctured by some inactivity 😄 👍

codeboy5:
I was previously using [ARFFFiles.jl](https://github.com/cjdoris/ARFFFiles.jl). It was not working with the multivariate time series, so I used this [function](https://github.com/timeseriesAI/tsai/blob/1e52def9c01c6ffa1be47d21e8cf163748e5e50a/tsai/data/external.py#L710) from tsai with some modifications. We are able to now read .ts for both univariate and multivariate time series.

PR #200, opened 2022-03-05
Adding Text ( NLP ) basic blocks => Initial stage

arcAman07:
Registered the NLP ( Text ) dataset to be added in the upcoming months. Added functions for the blocks of the Text dataset.
All the nlp dataset ( which are registered ) along with their forthcoming models will be added . Exploring Julia Text,  MLutils  and other package along with FastAI concepts so that these datasets can work well with Flux. As almost all the text datasets are in csv format it will be easily lo load them and create the corresponding container, working on further concepts to implement these text datasets.


arcAman07:
Currently I have added the entire basic structure of the Text Data comprising of the blocks and the containers. Have researched a lot since a week ( understanding FastAI docs and codebase ). Currently working on adding textrow block along with the recipes.jl.
Also currently working on two datasets "imdb" and "amazon_review_full" as both have different folder structure so different blocks would be required. Also going through the 2 papers which have built state of the art model for these two datasets and working on its implementation.  Any reviews thus far will be appreciated,.

PR #202, opened 2022-03-09
Add Flux 0.13 compatibility

### LINKED ISSUE #201: Update for Flux 0.13
Flux will stop exporting `params`, which I think causes this error in tests here:
```
getbatch: Error During Test at /home/runner/.julia/packages/ReTest/WnRIG/src/ReTest.jl:517
[1118](https://github.com/FluxML/Flux.jl/runs/5451760234?check_suite_focus=true#step:6:1118)
  Got exception outside of a @test
[1119](https://github.com/FluxML/Flux.jl/runs/5451760234?check_suite_focus=true#step:6:1119)
  UndefVarError: params not defined
[1120](https://github.com/FluxML/Flux.jl/runs/5451760234?check_suite_focus=true#step:6:1120)
  Stacktrace:
[1121](https://github.com/FluxML/Flux.jl/runs/5451760234?check_suite_focus=true#step:6:1121)
    [1] paramsrec(m::Function)
[1122](https://github.com/FluxML/Flux.jl/runs/5451760234?check_suite_focus=true#step:6:1122)
      @ FluxTraining ~/.julia/packages/FluxTraining/PJMOa/src/learner.jl:140
[1123](https://github.com/FluxML/Flux.jl/runs/5451760234?check_suite_focus=true#step:6:1123)
```

lorenzoh:
Closes #201.

Depends on Flux 0.13 release and https://github.com/FluxML/FluxTraining.jl/pull/103


lorenzoh:
Apparently this fails on 0.12 due to changes necessary to `Embedding`s for 0.13. @darsnack do you know what could have caused this and if this means that 0.13 changes to `Embedding` are not forward compatible?

darsnack:
This isn't because of `Embedding`. `Parallel(connection, layer)` (i.e. a single branch) doesn't wrap the single layer in a tuple. So instead of iterating the tuple as expected, we are iterating the branch itself.

darsnack:
Okay this is cause https://github.com/FluxML/Flux.jl/pull/1862 added a type restriction to `Parallel` that should have probably always been there.

You could either remove 0.12 from the compat, special case 0.12 in the source, or on L50 of `models.jl`, you can do `Tuple(catbackbone[2].layers)` in the `mapreduce`.

lorenzoh:
Thanks! I'll try the forward-compatible fix so that FastAI.jl can be used with both 0.12 and 0.13

Issue #196, opened 2022-03-02
MLUtils.jl transition


lorenzoh:
In the medium-term, FastAI.jl will move to depend on MLUtils.jl instead of MLDataPattern.jl + LearnBase.jl + DataLoaders.jl.

## Data container transformations

`groupobs`, `joinobs` and the likes in `transformations.jl`. These have already been copied over to MLUtils.jl in https://github.com/JuliaML/MLUtils.jl/pull/21. Functions here can be removed once FastAI.jl depends on MLUtils.jl.

## `DataLoader`

Path to replace dependency on DataLoaders.jl. Support for parallel data iterators has been merged into MLUtils.jl (https://github.com/JuliaML/MLUtils.jl/pull/33), but is not released (yet) and since it uses a different task scheduler than DataLoaders.jl, possible performance regressions on full training workloads still need to be benchmarked.

If that is not an issue and MLUtils.jl releases a stable `DataLoader` with feature parity of DataLoaders.jl, the dependency on DataLoaders.jl will be dropped.

PR #207, opened 2022-03-26
Add Container and Block for Text

Chandu-4444:
Tried starting at creating a simple textual recipe based on ImageFolders dataset recipe. This specifically works for `imdb` and similar datasets. Any feedback is highly appreciated.


Chandu-4444:
```julia
julia> using FastAI

julia> name, recipe = finddatasets(blocks=(Any, Any), name="imdb")[1]
Pair{String, FastAI.Datasets.DatasetRecipe}("imdb", TextFolders(FastAI.Datasets.parentname, false, FastAI.Text.var"#2#4"()))

julia> data, blocks = loadrecipe(recipe, datasetpath("imdb"))
((mapobs(loadfile, ["/home/luna/.julia/datadeps/fastai-imdb/imdb/test/neg/0_2.txt", "/home/luna/.ju…]), mapobs(parentname, ["/home/luna/.julia/datadeps/fastai-imdb/imdb/test/neg/0_2.txt", "/home/luna/.ju…])), (TextBlock(), Label{String}(["neg", "pos"])))

julia> text, class = obs = getobs(data, 1000)
("Every movie I have PPV'd because Leonard Maltin praised it to the skies has blown chunks! 
Every single one! 
When will I ever learn?<br /><br />Evie is a raving Old Bag who thinks nothing of saying she's dying of breast cancer to get her way! 
Laura is an insufferable Medusa filled with  The Holy Spirit (and her hubby's protégé)! 
Caught between these harpies is Medusa's dumb-as-a-rock boy who has been pressed into weed-pulling servitude by The Old Bag!<br /><br />
As I said, when will I ever learn?<br /><br />
I was temporarily lifted out of my malaise when The Old Bag stuck her head in a sink, but, unfortunately, she did not die. 
I was temporarily lifted out of my malaise again when Medusa got mowed down, but, unfortunately, she did not die. 
It should be a capital offense to torture audiences like this!<br /><br />
Without Harry Potter to kick him around, Rupert Grint is just a pair of big blue eyes that practically bulge out of its sockets.  
Julie Walters's scenery-chewing (especially the scene when she \"plays\" God) is even more shameless than her character.
<br /><br />
At least this Harold bangs some bimbo instead of Maude. 
For that, I am truly grateful. And if you're reading this Mr. Maltin, you owe me \$3.99!", "neg")


```

Chandu-4444:
I started adding various text transforms as in fastai text module.

Chandu-4444:
I have started adding functions for replacing words that starts with uppercase letters, contains all uppercase letters with special tokens like xxup, xxmaj etc.

Chandu-4444:
Each observation now has xxbos special token to indicate the start of the sentence just like fastai do.

src/Text/blocks/text.jl
```diff
@@ -0,0 +1 @@
+struct TextBlock <: Block end
```
lorenzoh, line 1: Some more documentation on what this represents would be great. Is this a sentence, a paragraph? A `checkblock` is also needed. The convention is not to put `Block` in the name of `Block`s, so depending on what this represents, we may be able to find a more precise term.

What tasks does this adress? Is there a need for different text-related blocks to solve other tasks? I'm. not super familiar with deep NLP, but these questions may help making this more precise.

src/Text/Text.jl
```diff
@@ -0,0 +1,35 @@
...
+include("transform.jl")
+
+function __init__()
+    _registerrecipes()
+    @require Makie="ee78f7c6-11fb-53f2-987a-cfe4a2b5a57a" begin
+        import .Makie
+        import .Makie: @recipe, @lift
+        import .FastAI: ShowMakie
+        include("makie.jl")
+    end
```
lorenzoh, line 29: We can drop this until we need to add Makie.jl recipes

src/Text/recipes.jl
```diff
@@ -0,0 +1,45 @@
+"""
+    TextFolders(textfile; labelfn = parentname, split = false)
+
+Recipe for loading a single-label text classification dataset
+stored in hierarchical folder format. 
+"""
+
```
lorenzoh, line 7:
```suggestion
```

src/Text/Text.jl
```diff
@@ -0,0 +1,35 @@
...
+end
+
+export TextBlock, TextFolders, replace_all_caps, replace_sentence_case,
+    convert_lowercase
```
lorenzoh, line 33: We should hold off on exporting everything except the block
```suggestion
export TextBlock
```

lorenzoh (changes requested):
Thanks for the PR!

I've left some comments.

Is there a fastai tutorial that uses this dataset? Would be helpful to know what kind of tasks could be tackled with this.

Chandu-4444:
> Is there a fastai tutorial that uses this dataset? Would be helpful to know what kind of tasks could be tackled with this.

Yes, fastai does have a tutorial that uses this dataset (And it's the first tutorial that focuses on the text module of the fastai library), https://docs.fast.ai/tutorial.text.html. This tutorial focuses on the sentiment analysis based on the given review. This first part uses a pre-trained language model (called [AWD-LSTM](https://arxiv.org/abs/1708.02182)) on Wikipedia for predicting the next word (language generation), and is directly used for predicting the sentiment for the given review. In the second part of the tutorial, they used a [ULMFit](https://arxiv.org/abs/1801.06146) approach that involves fine-tuning the model with the IMDB dataset and using that for predicting the sentiment. They achieved SOTA using the second method. 

I'll commit to the suggestions provided above and will improve this.  

Simultaneously, I'll start looking into that AWD-LSTM (https://arxiv.org/abs/1708.02182) paper to get deeper into how the model works. After that, the plan was to go through the ULMFit (https://arxiv.org/abs/1801.06146) paper.

src/Text/Text.jl
Chandu-4444: Ok. I'll remove this.

src/Text/blocks/text.jl
Chandu-4444: > Some more documentation on what this represents would be great. Is this a sentence, a paragraph?

The initial idea was to replicate `struct Image{N} <: Block end`. I haven't specified the dimension `N` as I'm not sure how to incorporate it with textual data is it the length? Can that be the vocabulary size? Or can we just leave this as it is? (Maybe we can discuss this). 

> Is there a need for different text-related blocks to solve other tasks?

Most text-related tasks revolve around classification and sentiment analysis (Which are single label prediction tasks). There are some more tasks that involve predicting multiple labels like topic modeling (Need more than one label to be predicted). Finally, there are some unsupervised techniques like text generation, text summarization.

I guess all the observations for the above tasks can be represented using a block of the same type?

> Is this a sentence, a paragraph?

I think the better name would be `Paragraph` rather than `TextBlock` as most if not all of the text data has a single paragraph (multiple sentences) as an individual observation (We can consider a sentence as a paragraph too!).

Chandu-4444: >  A checkblock is also needed.

I tried adding the `checkblock` for `TextBlock`, but it is always returning false. I guess it is calling the default `checkblock` method. I'm sure why this happens. Also the method I wrote isn't even getting displayed in the output of `methods(checkblock)`.

src/Text/recipes.jl
```diff
@@ -0,0 +1,44 @@
...
+function Datasets.loadrecipe(recipe::TextFolders, path)
+    isdir(path) || error("$path is not a directory")
+    data = loadfolderdata(
+        path,
+        filterfn=f -> istextfile(f) && recipe.filefilterfn(f),
+        loadfn=(loadfile, recipe.labelfn),
+        splitfn=recipe.split ? grandparentname : nothing)
+
+    (recipe.split ? length(data) > 0 : nobs(data) > 0) || error("No text files found in $path")
```
ToucheSir, line 23: Doesn't `nobs` fall back to `length` now (or vice versa)? Trying to understand why the nested ternary is necessary.

src/Text/recipes.jl
```diff
@@ -0,0 +1,44 @@
...
+
+    labels = recipe.split ? first(values(data))[2] : data[2]
+    blocks = TextBlock(), Label(unique(eachobs(labels)))
```
ToucheSir, line 26: Was this meant to be 
```suggestion
    blocks = (TextBlock(), Label(unique(eachobs(labels))))
```
? Given the lines below, I'd recommend declaring them separately and doing the tuple packing in the return.

src/Text/recipes.jl
```diff
@@ -0,0 +1,44 @@
...
+    length(blocks[2].classes) > 1 || error("Expected multiple different labels, got: $(blocks[2].classes))")
+    return data, blocks
+end
+
+# Registering recipes
+
+const RECIPES = Dict{String,Vector{Datasets.DatasetRecipe}}(
+    "imdb" => [TextFolders(
+        filefilterfn = f->!contains(f, "tmp_clas") && !contains(f, "tmp_lm") && !contains(f, "unsup")
```
ToucheSir, line 35:
```suggestion
        filefilterfn = f -> !occursin(f, "tmp_clas|tmp_lm|unsup")
```
A comment about why directories with these file names are to be avoided would be good too.

src/Text/transform.jl
```diff
@@ -0,0 +1,25 @@
...
+"""
+
+function replace_sentence_case(t::String)
+    t = replace(t, r"(?<!\w)([A-Z][A-Z0-9]*[a-z0-9]+)(?!\w)" => s"xxmaj \1")
+    t = replace(t, r"(?<!\w)([A-Z][A-Z0-9]*[a-z0-9]+)(?!\w)" => lowercase)
+end
+
+function convert_lowercase(t::String)
+    string("xxbos ", lowercase(t))
+end
```
ToucheSir, line 25:
```suggestion
function replace_all_caps(t::String)
    t = replace(t, r"([A-Z]+[^a-z\s]*)(?=(\s|$))" => s"xxup \1")
    return replace(t, r"([A-Z]*[^a-z\s]+)(?=(\s|$))" => lowercase)
end

"""
    replace_sentence_case(String)

Replace tokens in Sentence Case by their lower verions and add xxmaj before.
"""
function replace_sentence_case(t::String)
    t = replace(t, r"(?<!\w)([A-Z][A-Z0-9]*[a-z0-9]+)(?!\w)" => s"xxmaj \1")
    return replace(t, r"(?<!\w)([A-Z][A-Z0-9]*[a-z0-9]+)(?!\w)" => lowercase)
end

convert_lowercase(t::String) = string("xxbos ", lowercase(t))

```
Not sure if the type annotations are necessary.

src/datasets/containers.jl
```diff
@@ -38,6 +38,8 @@ function loadfile(file::String)
         return FileIO.load(file, view = true)
     elseif endswith(file, ".csv")
         return DataFrame(CSV.File(file))
+    elseif endswith(file, ".txt")
+        return String(read(file))
```
ToucheSir, line 42:
```suggestion
        return read(file, String)
```
Per last signature in https://docs.julialang.org/en/v1/base/io-network/#Base.read.

src/Text/recipes.jl
Chandu-4444: Including data from `unsup` folder for unsupervised text classification can be done in the later stages. And the other two directories were just some metadata.

src/Text/recipes.jl
Chandu-4444: Here, If the split is set to true then it checks for the corresponding loadfile mapobs and parentname mapobs by using `length(data)` which returns 2 when we have a valid `data` when using split. The next case would be checking for total obs in the data, done by `nobs(data)` return 50000 in this case. If none of this happens, we throw an error.

This is how I've seen it was implemented in `Vision/recipes.jl`.

src/Text/recipes.jl
Chandu-4444: Even this is how I've seen it was implemented in `Vision/recipes.jl`. But if it seems reasonable to do this way, we can change it.

src/Text/recipes.jl
ToucheSir: Got it. I think there needs to be a bit more documentation on this across recipes (or at least more well-named things in code so that one can figure out what all the `[2]` business is about, but that need not concern this PR.

lorenzoh: FastAI.jl does not use MLUtils.jl yet, see #196. Yes for the docs; in short, `blocks` is a tuple of `(inputblock, targetblock)` so `blocks[2]` takes the target block.

src/Text/blocks/text.jl
lorenzoh: > The initial idea was to replicate struct Image{N} <: Block end. I haven't specified the dimension N as I'm not sure on incorporating it with textual data. Is it the length? Can that be the vocabulary size? Or can we just leave this as it is? (Maybe we can discuss this).

Don't think this is necessary. As described in the documentation of `Block`, the block should only hold information that is constant for all observations, so encoding vocab size here would mean we assume it is fixed for all observations in a dataset which is too restrictive.

> I guess all the observations for the above tasks can be represented using a block of the same type?

Yeah, that sounds like it 👍 

> I think the better name would be Paragraph rather than TextBlock as most if not all of the text data has a single paragraph (multiple sentences) as an individual observation (We can consider a sentence as a paragraph too!).

`Paragraph` is good :)

lorenzoh (changes requested):
Left a comment regarding the block type, but other than that and Brian's open suggestion, this looks good to me!

src/Text/blocks/text.jl
Chandu-4444: > Paragraph is good :)

Yes, that's what I thought would be more reasonable. Changed it now.

lorenzoh:
Sorry for letting this sit!

Tests were failing due to issues that should be fixed on master, so merging master into this should make the CI green.

Last thing that would be good to have would be some tests

Chandu-4444:
Sure! Will synchronise it with master and add some tests.

Chandu-4444:
Umm... For writing tests to the TextFolders(), I need to access the IMDb dataset. I remember Lorenz mentioning that it isn't very nice to use large datasets for testing as it might overload the CI system. And for other recipes, there are smaller version datasets that replicate the original larger version datasets. I couldn't find any such datasets for IMDb (Actually there is one such dataset that is available as a CSV file, but I need an IMDb-like directory structure for testing the recipe). Is there any workaround?

ToucheSir:
I wouldn't worry about testing the bits that require file IO for now, mostly the helper functionality.

Chandu-4444:
That sounds good!

src/Textual/blocks/text.jl
```diff
@@ -26,4 +26,4 @@ FastAI.mockblock(Paragraph())
 struct Paragraph <: Block end
 
 FastAI.checkblock(::Paragraph, ::String) = true
-FastAI.mockblock(::Paragraph) = randstring(" ABCEEFGHIJKLMNOPQESRUVWXYZ 1234567890abcdefghijklmnopqrstynwxyz\n\t.,", 40)
+FastAI.mockblock(::Paragraph) = randstring(" ABCEEFGHIJKLMNOPQESRUVWXYZ 1234567890 abcdefghijklmnopqrstynwxyz\n\t.,", 40)
```
lorenzoh, line 29: `Paragraph` can have variable length, right? Let's put this in `mockblock` so that we don't bake in invariants that don't need to hold: 
```suggestion
FastAI.mockblock(::Paragraph) = randstring(" ABCEEFGHIJKLMNOPQESRUVWXYZ 1234567890 abcdefghijklmnopqrstynwxyz\n\t.,", rand(10:40))
```

src/Textual/recipes.jl
```diff
@@ -0,0 +1,51 @@
...
+        Datasets.registerrecipe!(Datasets.FASTAI_DATA_REGISTRY, name, recipe)
+    end
+end
+
+
+## Tests
+
+
+@testset "TextFolders [Recipe]" begin
+    @test length(Datasets.FASTAI_DATA_REGISTRY.recipes["imdb"]) == 1
```
lorenzoh, line 50: This is brittle in that it will break when a new recipe is added to `"imdb"` and goes into the internals of the dataset registry. This should test `finddatasets("imdb", B)` returns _at least one_ recipe for some block `B` we want to test.

src/Textual/transform.jl
```diff
@@ -0,0 +1,22 @@
...
+
+Replace tokens in ALL CAPS by their lower version and add xxup before.
+"""
+
+function replace_all_caps(t)
+    t = replace(t, r"([A-Z]+[^a-z\s]*)(?=(\s|$))" => s"xxup \1")
+    return replace(t, r"([A-Z]*[^a-z\s]+)(?=(\s|$))" => lowercase)
+end
+
+"""
```
lorenzoh, line 12: Some inline tests for these preprocessing steps would be good to have, too

lorenzoh (changes requested):

src/Textual/blocks/text.jl
Chandu-4444: Yeah, that's more reasonable thing to do.

src/Textual/transform.jl
Chandu-4444: Yeah, that's in my to-do list. Just wanted to make sure they're necessary. I'll add some tests in my next commit.

src/Textual/recipes.jl
Chandu-4444: Sure. I'll update the test.

PR #219, opened 2022-04-20
add ImageIO backend

### LINKED ISSUE #216: Use JpegTurbo.jl to load .jpg images
JpegTurbo.jl can speed up image loading quite a bit, and the previous ImageIO.jl versions had severe performance regressions for 1.7, which should be fixed if JpegTurbo.jl is used.

johnnychen94:
closes #216


johnnychen94:
@lorenzoh the test dependency Makie needs an update to resolve the compatibility.

PR #221, opened 2022-04-27
Keypoint regression example

itan1:
Running these steps in my REPL worked but only gave me the following problem
- the `FastAI.decodeypred(task, Training(), y)` check does not have the same coordinates as the original keypoint `k`

Running it in a local notebook gives me the following problems
- The first image plotted with `DataAugmentation.showitems` looks like this
Screenshot 2022-04-27 at 09 27 02
 - The kernel dies every time I try to run `showbatch(task, (xs, ys))`

Can someone try to reproduce this? I'm not sure whether these are environments problems (running with julia 1.7.2 and the mac M1 arm architecture is not fully supported yet).

### DIFF
--- a/notebooks/keypointregression.ipynb
+++ b/notebooks/keypointregression.ipynb
@@ -413,14 +413,7 @@
     }
    ],
    "source": [
-    "DLPipelines.decodeŷ(task, Training(), y)"
-   ]
-  },
-  {
-   "cell_type": "markdown",
-   "metadata": {},
-   "source": [
-    "We should also visualize our data to make sure that after all the encoding it still makes sense and the keypoint is properly aligned with the head on the image. The visualizations are derived from the data blocks we used when defining our `BlockTask`."
+    "FastAI.decodeypred(task, Training(), y)"
    ]
   },
   {
--- a/notebooks/keypointregression.ipynb
+++ b/notebooks/keypointregression.ipynb
@@ -18,22 +18,12 @@
    "cell_type": "code",
    "execution_count": 1,
    "metadata": {},
-   "outputs": [
-    {
-     "name": "stderr",
-     "output_type": "stream",
-     "text": [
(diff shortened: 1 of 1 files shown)


lorenzoh:
Thanks for taking a look at this!

> The first image plotted with DataAugmentation.showitems looks like this

I think this should be solved by a an `using ImageShow` import

> the FastAI.decodeypred(task, Training(), y) check does not have the same coordinates as the original keypoint k

It does not give the same keypoint since the image is cropped to 128x128. Important part is that the decoded keypoint is within those bounds 👍 

> The kernel dies every time I try to run showbatch(task, (xs, ys))

Cannot reproduce this, may be a problem with Makie.jl on ARM :/ . Have you tried updating packages in your environment?

notebooks/keypointregression.ipynb
```diff
@@ -18,22 +18,12 @@
...
-      "└ @ Base loading.jl:1342\n",
-      "WARNING: using Makie.Label in module FastAI conflicts with an existing identifier.\n"
-     ]
-    }
-   ],
+   "outputs": [],
    "source": [
     "import CairoMakie; CairoMakie.activate!(type=\"png\")\n",
     "using DelimitedFiles: readdlm\n",
     "using FastAI\n",
```
lorenzoh, line 25:
```suggestion
    "using FastAI, ImageShow\n",
```

lorenzoh:
Changes look good to me, merging!

itan1:
Great thank you too! This package is helping me a lot 🙌 

> > the FastAI.decodeypred(task, Training(), y) check does not have the same coordinates as the original keypoint k
> 
> It does not give the same keypoint since the image is cropped to 128x128. Important part is that the decoded keypoint is within those bounds 👍

Ok thanks for the clarification, I think I got it. The decode function only reverses the coordinate normalization, not the cropping. Is that correct?

lorenzoh:
Exactly, since the cropping is a destructive operation

PR #226, opened 2022-05-14
Update TagBot.yml

lorenzoh:
This was using the outdated format. Now uses the newer format, as outlined here: https://discourse.julialang.org/t/ann-required-updates-to-tagbot-yml/49249

@darsnack I saw you've been manually tagging releases for FastAI.jl; this should fix the issue so no manual involvement is needed, right?

### DIFF
--- a/.github/workflows/TagBot.yml
+++ b/.github/workflows/TagBot.yml
@@ -1,9 +1,12 @@
 name: TagBot
 on:
-  schedule:
-    - cron: 0 0 * * *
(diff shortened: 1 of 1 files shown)


darsnack:
I think it uses my name but I haven't been doing anything manual. But yeah we should update to the latest version.

lorenzoh:
> I think it uses my name but I haven't been doing anything manual. 
Ah alright 😅

PR #239, opened 2022-06-18
Add Time Series Block

codeboy5:
Added Time Series Container and Block. It is capable of loading all datasets from [timeseriesclassification](http://timeseriesclassification.com/dataset.php). The `.ts` files are loaded using Julia translation of this [method](https://github.com/timeseriesAI/tsai/blob/600a5ee7efe04c6ff19c7a27f596f947c75f31ef/tsai/data/external.py#L1389) .

I have also added a basic test case for the recipe. 
This allows us to do the following 

```julia
using FastAI

data, blocks = load(datarecipes()["ecg5000"])
nobs(data)
sample = series, class = getobs(data, 10)
```

Just wanted to get some initial thoughts on the work, there might be more changes as I continue to work on the other parts.


codeboy5:
Hey for downloading the datasets, the message as well as the post fetch method needs to be different which I think might require some restructuring, how should I go about that ?

src/datasets/load.jl
```diff
@@ -1,42 +1,42 @@
...
 """
-    datasetpath(name)
+datasetpath(name)
 
 Return the folder that dataset `name` is stored.
 
 If it hasn't been downloaded yet, you will be asked if you want to
 download it. See [`Datasets.DATASETS`](#) for a list of available datasets.
 """
 function datasetpath(name)
```
ToucheSir, line 10: Looks like the indentation got a little messed up in this file?

src/datasets/load.jl
```diff
@@ -47,17 +47,18 @@ const RE_IMAGEFILE = r".*\.(gif|jpe?g|tiff?|png|webp|bmp)$"i
 isimagefile(f) = matches(RE_IMAGEFILE, f)
 const RE_TEXTFILE = r".*\.(txt|csv|json|md|html?|xml|yaml|toml)$"i
 istextfile(f) = matches(RE_TEXTFILE, f)
-
+const RE_TIMESERIESFILE = r".*\.(ts)$"i
```
ToucheSir, line 50: We'll have to figure out how to deal with time series datasets in different formats eventually. This may involve changing the text file detection logic. No need to worry about that now, this is just a note for when that time comes :)

src/datasets/containers.jl
```diff
@@ -94,6 +308,21 @@ LearnBase.getobs(dataset::TableDataset{<:CSV.File}, idx) = dataset.table[idx]
 LearnBase.nobs(dataset::TableDataset{<:CSV.File}) = length(dataset.table)
 
 
+#TimeSeriesDataset
+
+struct TimeSeriesDataset{}
+    table::AbstractArray{Float64,3}
```
ToucheSir, line 314:
```suggestion
struct TimeSeriesDataset{T<:AbstractArray}
    table::T
```
The type parameter is required for avoid an abstract field type. I've loosened the array type bound as well, but it may need to be tuned up or down later.

src/datasets/containers.jl
```diff
@@ -26,6 +26,218 @@ function rglob(filepattern = "*", dir = pwd(), depth = 4)
...
+
+                    if tokens[2] == "true"
+                        timestamps = true
+                    else
+                        timestamps = false
+                    end
+
+                    has_timestamps_tag = true
+                    metadata_started = true
+
```
ToucheSir, line 89: There should be a way to extract out some of this logic. For now, having tests would help verify it's working.

codeboy5: Yeah I think it could be extracted out for the current datasets we are using. I just copied it over from python library to get it working. 
By tests do you mean automated tests and creating a sample `.ts` file to run the test on ?

src/TimeSeries/blocks/timeseriesrow.jl
```diff
@@ -0,0 +1,37 @@
...
+
+{cell=main}
+```julia
+using FastAI
+FastAI.mockblock(TimeSeriesRow{1,10}())
+```
+
+"""
+
+struct TimeSeriesRow{M,N} <: Block end
```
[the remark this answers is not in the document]
codeboy5: - For the current datasets we are planning to use, the length is same for all observations, so we can have a constant block.

From my understanding we build the model from the block ? So the parameters might depend on time series length and number of variables. 
So ideally the block should just dispatch on parameters which are required in model building ?

src/TimeSeries/recipes.jl
```diff
@@ -0,0 +1,63 @@
...
+                recipe = recipe,
+            ))
+        end
+    end
+end
+
+# ## Tests
+
+@testset "TimeSeriesDataset [recipe]" begin
+    path = datasetpath("ecg5000")
```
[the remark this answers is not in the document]
codeboy5: The dataset is around 10mb, should we run this on CI ?

src/datasets/containers.jl
```diff
@@ -94,6 +308,21 @@ LearnBase.getobs(dataset::TableDataset{<:CSV.File}, idx) = dataset.table[idx]
 LearnBase.nobs(dataset::TableDataset{<:CSV.File}) = length(dataset.table)
 
 
+#TimeSeriesDataset
+
+struct TimeSeriesDataset{}
+    table::AbstractArray{Float64,3}
+end 
+
+function LearnBase.getobs(dataset::TimeSeriesDataset, idx)
```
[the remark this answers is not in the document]
codeboy5: Sure, will update it.

src/TimeSeries/recipes.jl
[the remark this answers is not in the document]
lorenzoh: That should be fine 👍

src/datasets/load.jl
codeboy5: Just to clarify, is there a way to auto format the code ?

src/datasets/containers.jl
ToucheSir: Correct. You could also use a small existing one as long as the license is compatible.

codeboy5:
I have implemented the changes we discussed in the last call.

- [Replace type parameters with fields](https://github.com/FluxML/FastAI.jl/blob/9117ac6de2ceaa4d590186d3e990c1f0dd803419/src/TimeSeries/blocks/timeseriesrow.jl#L26)
- [TSClassificationDatasetStruct](https://github.com/FluxML/FastAI.jl/blob/9117ac6de2ceaa4d590186d3e990c1f0dd803419/src/datasets/fastaidatasets.jl#L12) to load the `.ts` files from http://timeseriesclassification.com .
- Test for `.ts2df` function as @ToucheSir suggested. The size for file is 159KB, should I just include it in the commit or download It on the fly. As for the license, according to [tsml_repo](https://github.com/uea-machine-learning/tsml_repo), it seems to have a [GNU General Public License v3.0](https://github.com/uea-machine-learning/tsml_repo/blob/master/LICENSE) .

ToucheSir:
Great. Just to be safe, I think it would be best to download the file on the fly for now.

src/TimeSeries/encodings/timeseriespreprocessing.jl
```diff
@@ -0,0 +1,51 @@
...
+    end 
+    axes = [ax for ax in [1, 2, 3] if !(ax in drop_axes)]
+    mean = Statistics.mean(data, dims=axes)
+    std  = Statistics.std(data, dims=axes)
+    return mean, std
+end
+
+function setup(::Type{TSPreprocessing}, ::TimeSeriesRow, data)
+    means, stds = tsdatasetstats(data)
+end
```
lorenzoh, line 51:
```suggestion
setup(::Type{TSPreprocessing}, ::TimeSeriesRow, data) = means, stds = tsdatasetstats(data)
```

src/TimeSeries/TimeSeries.jl
```diff
@@ -31,12 +31,22 @@ using InlineTest
...
+        if !haskey(FastAI.learningtasks(), t.id)
+            push!(FastAI.learningtasks(), t)
+        end
+    end
 end
 
-export TimeSeriesRow
-
+export 
+    TimeSeriesRow, TSClassificationSingle, TimeSeriesPreprocessing
```
lorenzoh, line 51:
```suggestion
export 
    TimeSeriesRow, TSClassificationSingle, TSPreprocessing
```

src/TimeSeries/encodings/timeseriespreprocessing.jl
```diff
@@ -0,0 +1,51 @@
...
+
+struct TSPreprocessing <: Encoding
+    tfms
+end
+
+function TSPreprocessing()
+    base_tfms = [
+    ]
+    return TSPreprocessing(base_tfms)
+end
```
lorenzoh, line 19: What kind of transforms will be in here?

src/TimeSeries/encodings/timeseriespreprocessing.jl
```diff
@@ -0,0 +1,51 @@
...
+
+function encodedblock(p::TSPreprocessing, block::TimeSeriesRow)
+    return block
+end
```
lorenzoh, line 23: If the format of the time series is changed by the encoding, this should return a different block

codeboy5: No the format won't be changed, as I discussed with Brian earlier that different models might require different formats and so the encoding shouldn't depend on the model.

codeboy5:
I have added the basic structure and code. There seems to be a couple of errors, I will solve by tonight. 
Let me know how does the basic code look.

src/TimeSeries/encodings/timeseriespreprocessing.jl
```diff
@@ -1,3 +1,12 @@
+struct TSStats{T}
+    means::AbstractArray{T,2}
+    stds::AbstractArray{T,2}
+end
+
+function TSStats(means, stds)
+    TSStats{eltype(means)}(means, stds)
+end
```
lorenzoh, line 8: Not sure we need this struct, it may be simpler to add `means` and `stds` fields to the `Encoding`. Then that also makes it easier to construct `TSPreprocessing` manually (i.e. without `setup`).

src/TimeSeries/encodings/timeseriespreprocessing.jl
```diff
@@ -10,23 +19,25 @@ Encodes
...
-    return block
+    return TSPreprocessing{}(base_tfms, stats)
 end
 
-function encode(p::TSPreprocessing, context, block::TimeSeriesRow, obs)
-    for tfm in values(p.tfms)
-        obs = tfm(obs)
-    end
-    obs
+function TSStandardize(
```
lorenzoh, line 32: Should this be an `encode` method?

src/TimeSeries/encodings/timeseriespreprocessing.jl
```diff
@@ -41,11 +52,26 @@ function tsdatasetstats(
...
+    stds  = reshape(stds, ( size( stds)[2:3] ))
+    stats = TSStats(means, stds)
+    return TSPreprocessing(stats)
+end
+
+function encodedblock(p::TSPreprocessing, block::TimeSeriesRow)
+    return block
+end
+
+function encode(p::TSPreprocessing, context, block::TimeSeriesRow, obs)
```
lorenzoh, line 72: Can we add a check of the dimensions of `obs` against the means/stds here?

lorenzoh (commented):
Left some minor comments, but let me know when you're done with the other changes :) :+1:

src/TimeSeries/blocks/timeseriesrow.jl
```diff
@@ -0,0 +1,42 @@
...
+"""
+
+struct TimeSeriesRow <: Block 
+    nfeatures::Int
+    obslength::Union{Int, Colon}
+end
+
+function checkblock(row::TimeSeriesRow, obs::AbstractArray{T,2}) where {T<:Number}
+    size(obs) == (row.nfeatures, row.obslength)
+end
```
ToucheSir, line 33:
```suggestion
checkblock(row::TimeSeriesRow, obs::AbstractMatrix{<:Number}) = size(obs) == (row.nfeatures, row.obslength)
```

src/TimeSeries/encodings/timeseriespreprocessing.jl
```diff
@@ -0,0 +1,77 @@
...
+
+function TSStandardize(
+    obs,
+    p::TSPreprocessing
+)
+    means = p.stats.means
+    stds  = p.stats.stds
+    obs = obs .- means
+    obs = obs ./ stds
+    return obs
```
ToucheSir, line 40:
```suggestion
    return (obs .- p.stats.means) ./ p.stats.stds
```
This will fuse and save an intermediate allocation.

src/TimeSeries/encodings/timeseriespreprocessing.jl
```diff
@@ -0,0 +1,77 @@
...
+    by_var=false,
+    by_step=false
+)
+    drop_axes = []
+    if (by_var)
+        append!(drop_axes,2)
+    else
+        append!(drop_axes,3)
+    end 
+    axes = [ax for ax in [1, 2, 3] if !(ax in drop_axes)]
```
ToucheSir, line 54:
```suggestion
    drop_axis = by_var ? 2 : 3
    axes = filter(∉(drop_axis), 1:3)
```
It appears `drop_axes` only ever has one axis, so I've taken the liberty to simplify it here. Unless you were planning on having an if statement for `by_step` as well?

src/TimeSeries/encodings/timeseriespreprocessing.jl
```diff
@@ -0,0 +1,77 @@
...
+    mean = Statistics.mean(data.table, dims=axes)
+    std  = Statistics.std(data.table, dims=axes)
```
ToucheSir, line 56:
```suggestion
    std  = Statistics.std(data.table, dims=axes, mean=mean)
```
If the data source supports it, this is more efficient.

src/TimeSeries/encodings/timeseriespreprocessing.jl
```diff
@@ -0,0 +1,77 @@
...
+    return mean, std
+end
+
+function setup(::Type{TSPreprocessing}, ::TimeSeriesRow, data)
+    means, stds = tsdatasetstats(data[1])
+    means = reshape(means, ( size( means)[2:3] ))
+    stds  = reshape(stds, ( size( stds)[2:3] ))
```
ToucheSir, line 63: A little weird formatting going on here. Why is the reshape needed?

src/datasets/containers.jl
```diff
@@ -1,4 +1,216 @@
...
+        for sample in 1:num_samples
+            data[sample, 1:end, 1:end] = instance_list[sample]
+        end
+
+        # Check if we should return any associated class labels separately
+        if class_labels
+            return data, class_val_list
+        else
+            return data
+        end
```
ToucheSir, line 208: Returning one or two things based on a conditional is a little surprising. Consider either always returning class_val_list and/or making the second return value some meaningful null value (nothing, empty array, whatever makes the most sense).

src/TimeSeries/encodings/timeseriespreprocessing.jl
```diff
@@ -0,0 +1,77 @@
+struct TSStats{T}
+    means::AbstractArray{T,2}
+    stds::AbstractArray{T,2}
```
ToucheSir, line 3: You could also consider changing these to `AbstractMatrix{T}`, but it would be best to confirm first that they will always be 2-dimensional (what about higher-dimensional time series, for example?)

codeboy5:
Hey the encodings are working now.

```julia
> using FastAI
> data, blocks = load(datarecipes()["ecg5000"]);
> task = FastAI.TSClassificationSingle(blocks, data);
> input, target = getobs(data, 2)
(Float32[-1.1008778 -3.9968398 … 1.1196209 -1.4362499], "1")
> encodesample(task, Training(), (input, target))
(Float32[-1.1048299 -4.011188 … 1.1236403 -1.4414059], Float32[1.0, 0.0, 0.0, 0.0, 0.0])
```

I am looking to add some tests as well look to resolve all the comments you guys made before the meeting. Have also started working on creating the demo notebook side-by-side as you guys suggested.

src/TimeSeries/encodings/timeseriespreprocessing.jl
codeboy5: Yeah this would work, we can have the normalisation over sample too but yeah for now this would look better.

src/TimeSeries/blocks/timeseriesrow.jl
codeboy5: Yeah this looks better, thanks 👍

src/TimeSeries/encodings/timeseriespreprocessing.jl
codeboy5: Not sure about the higher dimensional series, I will have to check the literature or some examples online if I can find them.

ToucheSir: Oh I can come up with examples easily enough, the question is whether the fastai docs have any :)

codeboy5: Oh, I will check that up before our meeting today.

codeboy5: Oh, I will check that up before our meeting today.

src/TimeSeries/encodings/timeseriespreprocessing.jl
codeboy5: Yes, I tried using `Statistics.stdm(itr, mean)` earlier, couldn't seem to get it working. Will look into it further and keep this comment open till then.

codeboy5:
I've added the decode method and basic visualisation too and updated the notebook accordingly. The formatting for plots Is a little weird on the notebooks, but looks fine on the terminal. Do you guys know how to fix this, should I try adjusting the size of the plot ?

Screenshot 2022-07-06 at 10 07 39 AM

lorenzoh:
> The formatting for plots Is a little weird on the notebooks, but looks fine on the terminal. 

Rendering of Unicode plots can be a bit off depending on the notebook environment/viewer. It should end up looking fine in the docs, though.

lorenzoh:
Having some trouble to push the subpackage merge to your fork, @codeboy5. Can you check that I have permissions to push to your fork? Or, alternatively just give me committer access to the fork

image

codeboy5:
> Having some trouble to push the subpackage merge to your fork, @codeboy5. Can you check that I have permissions to push to your fork? Or, alternatively just give me committer access to the fork
> 
> image

If checked, users with write access to FluxML/FastAI.jl can add new commits to your timeseries-blocks branch. You can always change this setting later.
This option is checked out for me.
Should I add you as a collaborator on my forked repo ?

lorenzoh:
> > Having some trouble to push the subpackage merge to your fork, @codeboy5. Can you check that I have permissions to push to your fork? Or, alternatively just give me committer access to the fork
> > 
> > image
> 
> If checked, users with write access to FluxML/FastAI.jl can add new commits to your timeseries-blocks branch. You can always change this setting later.
> This option is checked out for me.
> Should I add you as a collaborator on my forked repo ?

Yeah, please try that

codeboy5:
> > > Having some trouble to push the subpackage merge to your fork, @codeboy5. Can you check that I have permissions to push to your fork? Or, alternatively just give me committer access to the fork
> > > image
> > 
> > 
> > If checked, users with write access to FluxML/FastAI.jl can add new commits to your timeseries-blocks branch. You can always change this setting later.
> > This option is checked out for me.
> > Should I add you as a collaborator on my forked repo ?
> 
> Yeah, please try that

Yeah, just sent you an invite for the same. Hopefully it should work now.

codeboy5:
With this commit, we are able to the following now 

```julia
> using FastAI, FastTimeSeries
> data, blocks = load(datarecipes()["ecg5000"]);
> task = FastTimeSeries.TSClassificationSingle(blocks, data);
> backbone = FastTimeSeries.Models.StackedLSTM(1, 16, 10, 2);
> model = FastAI.taskmodel(task, backbone)
Chain(
  StackedLSTMCell(
    Recur(
      LSTMCell(1 => 10),                # 500 parameters
    ),
    Recur(
      LSTMCell(10 => 16),               # 1_760 parameters
    ),
  ),
  identity,
  Dense(16 => 5),                       # 85 parameters
)         # Total: 12 trainable arrays, 2_345 parameters,
          # plus 4 non-trainable, 13_312 parameters, summarysize 62.145 KiB.
```

FastTimeSeries/src/models/StackedLSTM.jl
```diff
@@ -0,0 +1,50 @@
...
+					  LSTM(hiddensize, out; init=init))
+	else
+		chain_vec = [LSTM(in, hiddensize; init=init)]
+		for i = 1:layers - 2
+			push!(chain_vec, LSTM(hiddensize, hiddensize; init=init))
+		end
+		chain = Chain(chain_vec..., LSTM(hiddensize, out; init=init))
+	end
+	return StackedLSTMCell(chain)
+end
```
ToucheSir, line 34: Something is a little weird here: you're returning a `Cell`, but inside that cell are `Recur`s and not `LSTMCells`. Ideally `StackedLSTMCell` would be structured like the other Flux RNN cells, but if that's not possible I'd recommend renaming to `StackedLSTM` to better represent the model as stateful + containing internal mutation (transitively via `Recur` from `LSTM`).

FastTimeSeries/src/models/StackedLSTM.jl
```diff
@@ -0,0 +1,50 @@
+# StackedLSTM.jl
+#
+# Layers for stacked LSTM (referenced from https://github.com/sdobber/FluxArchitectures.jl)
+
+mutable struct StackedLSTMCell{A}
```
ToucheSir, line 5:
```suggestion
struct StackedLSTMCell{A}
```

FastTimeSeries/src/models/StackedLSTM.jl
```diff
@@ -0,0 +1,50 @@
...
+Flux.@functor StackedLSTMCell 
+Flux.trainable(m::StackedLSTMCell) = (m.chain)
+
+# Initialize forget gate bias with 1
+function initialize_bias!(l::StackedLSTMCell)
+	for i = 1:length(l.chain)
+		l.chain[i].cell.b .= 1
+	end
+	return nothing
+end
```
ToucheSir, line 50: This should be done explicitly through the `initb` argument of the LSTM(Cell) constructor if possible.

FastTimeSeries/src/models/StackedLSTM.jl
```diff
@@ -0,0 +1,50 @@
...
+	return StackedLSTMCell(chain)
+end
+
+function (m::StackedLSTMCell)(X)
+	[m.chain(x) for x ∈ X[1:end-1]]
+	return m.chain(X[end])
+end
+
+Flux.@functor StackedLSTMCell 
+Flux.trainable(m::StackedLSTMCell) = (m.chain)
```
ToucheSir, line 42:
```suggestion
```
This is redundant with the `@functor` above.

FastTimeSeries/src/models/StackedLSTM.jl
```diff
@@ -0,0 +1,50 @@
...
+			push!(chain_vec, LSTM(hiddensize, hiddensize; init=init))
+		end
+		chain = Chain(chain_vec..., LSTM(hiddensize, out; init=init))
+	end
+	return StackedLSTMCell(chain)
+end
+
+function (m::StackedLSTMCell)(X)
+	[m.chain(x) for x ∈ X[1:end-1]]
+	return m.chain(X[end])
```
ToucheSir, line 38: Is this always a sequence-to-one model?

FastTimeSeries/src/tasks/classification.jl
```diff
@@ -26,4 +26,33 @@ _tasks["tsclfsingle"] = (
...
+    @test_nowarn tasklossfn(task)
+    
+    @testset "`encodeinput`" begin 
+        row = FastAI.mockblock(task.blocks[1])
+
+        xtrain = FastAI.encodeinput(task, Training(), row);
+        @test length(xtrain[1,:]) == task.blocks.input.obslength
+        @test length(xtrain[:,1]) == task.blocks.input.nfeatures
+
+        @test eltype(xtrain[1,:]) <: Number
```
ToucheSir, line 48: Is the input expected to be non-contiguous here? If not, I would compare against `size` instead of lengths of slices.

FastTimeSeries/src/models/StackedLSTM.jl
codeboy5: I think for classification and regression, this will always be a seq-to-one model ?
For some nlp tasks, it would not be.

ToucheSir: Ok, then I think you should be able to rewrite this using `foldl`. It'll be more correct and likely perform better as well. It may also let you drop the `Recur` wrapper for the inner layer stack and use the cells directly instead (i.e. `LSTM -> LSTMCell`).

codeboy5:
image

FastTimeSeries/src/models/RNN.jl
```diff
@@ -1,3 +1,9 @@
+function tabular2rnn(X::AbstractArray{Float32, 3})
+    X = permutedims(X, (2, 1, 3))
+    X = [X[t, :, :] for t ∈ 1:size(X, 1)]
```
ToucheSir, line 3:
```suggestion
    X = collect(eachslice(X; dims=1))
```
Should be more efficient. But see comments below.

FastTimeSeries/src/models/RNN.jl
```diff
@@ -24,5 +30,5 @@ function RNNModel(recbackbone,
                   dropout_rate = 0.0)
     
     dropout = dropout_rate == 0 ? identity : Dropout(dropout_rate)
-    Chain(recbackbone, dropout, finalclassifier)
-end
+    Chain(tabular2rnn, recbackbone, dropout, finalclassifier)
```
ToucheSir, line 33: I don't think we should be doing this dense to slices transform in the model itself. If you just need something RNN friendly, relying on the built-in support for dense inputs should be enough. The `permutedims` can stay for now, but even that probably shouldn't be in the gradient hot path (it allocates O(n) both ways).

ToucheSir: An alternative for now is to make the data pipeline spit out the vector of arrays. We can then revisit if/when you add models like CNNs which expect dense inputs.

FastTimeSeries/src/models/RNN.jl
```diff
@@ -1,3 +1,9 @@
+function tabular2rnn(X::AbstractArray{Float32, 3})
```
ToucheSir, line 1: I think "tabular" is a bit of a misnomer here, but naming is not a priority.

FastTimeSeries/src/models/RNN.jl
codeboy5: I think the "ideal" place to do this transform would be inside the training loop. I am not sure how to exactly do that for FastAI.jl. Is there a way ?

Since the second phase would involve using some CNNs, using data pipeline to spit out vector of arrays would not work.

lorenzoh:
Since this PR is getting quite long, it may be best to move the newest changes to a new PR and merge the parts that are ready. We could revert the PR to the point where I did the manual merge, merge this PR with the changes up to that point into master, and then open a new PR based on master with the more recent changes, including models and so on. @codeboy5 let me know if you aren’t sure about how to do this 🙂

codeboy5:
> Since this PR is getting quite long, it may be best to move the newest changes to a new PR and merge the parts that are ready. We could revert the PR to the point where I did the manual merge, merge this PR with the changes up to that point into master, and then open a new PR based on master with the more recent changes, including models and so on. @codeboy5 let me know if you aren’t sure about how to do this 🙂

Yeah sure, just have to revert the code to the pr right ?
Will do it.

codeboy5:
@lorenzoh Does this look okay ?

On the last meeting, @darsnack suggested some changes to the model code, I will commit them in the next PR.

lorenzoh:
Yeah this looks good! Just to make sure, @darsnack does it make sense to rebase this on master when merging or should we squash and merge?

darsnack:
Either is okay. Maybe squash and merge will be less painful?

lorenzoh:
I've squashed and merged it. @codeboy5 you should be able to open a new PR for the model code based on master now.

codeboy5:
> I've squashed and merged it. @codeboy5 you should be able to open a new PR for the model code based on master now.

will do thanks . 👍🏻

PR #240, opened 2022-06-20
Move domain-specific functionality to subpackages

### LINKED ISSUE #224: Makie 0.17 support
When using Makie.jl 0.17, there are some errors cropping (see https://github.com/FluxML/FastAI.jl/runs/6404954707?check_suite_focus=true) up in the code for showing multiple observations, here: https://github.com/FluxML/FastAI.jl/blob/2f227aab66bdb554b6a628903a618a57dbc290c3/src/interpretation/makie/showmakie.jl#L54-L58

This code tries to add titles to the beginning of each column of a previously created `GridLayout`.

### LINKED ISSUE #241: Make a subpackage for Makie support
FastMakie can be a subpackage living in the FastAI repo that implements the Makie `ShowBackend`. Subpackages can then depend on MakieCore.jl to define recipes and we don't need Requires anymore.

Since this will be breaking as well, it should be released together with the other breaking changes #196 and #240. #224 also needs to be fixed for this.

lorenzoh:
This turns the domain-specific modules into packages as discussed in #186. Specifically:
- `FastAI.Vision` -> `FastVision`
- `FastAI.Tabular` -> `FastTabular`
- `FastAI.Textual` -> `FastText`
- `FastMakie` for Makie backend

This is obviously a gigantic breaking change and I'm thinking to tag this together with #196 to avoid subsequent breaking changes. Thoughts @darsnack? This should also be the last of the larger (and breaking) changes I have been planning to make, so if there are any other breakages planned, we should tag them together.

Closes #224, closes #241 

Status:

- [x] create subpackages
- [x] move and fix tests
- [x] add test projects for subpackages
- [x] create CI for subpackages
- [x] apply JuliaFormatter (SciMLStyle)
- [x] update documentation to work with subpackages
  - [x] update markdown sources and docstrings
  - [x] rerun notebooks


darsnack:
I think based on our experience with NNlib, the subpackage approach seems to be the best option moving forward. The only other possible breaking change I can think off is the data blocks API for time series support.

lorenzoh:
FastAI is now a core package, with FastVision etc. providing the domain-specifc functionality. So to run computer vision tasks, one imports `using FastAI, FastVision`. Loading the domain packages also populates the feature registries (`datasets`, `datarecipes` and `learningtasks`). With more domains being added, reducing the number of total packages that have to be loaded helps reduce the import time.

lorenzoh:
> The only other possible breaking change I can think off is the data blocks API for time series support.

What would be breaking about that?

darsnack:
Maybe this has been resolved in more recent discussions. I recall that we mentioned how for sequential data, each sample is a sequence of features. So the outermost list is over samples like any dataset. But for Flux models, the outermost list is over time steps, and each element is a batch of features at a given time step. This means that there is an "encoding" (i.e. `MLUtils.bathseq`) that needs to happen _after_ batching. It's definitely a solvable issue, but I didn't know how you wanted to handle it. Potentially the solution would be breaking or non-breaking.

PR #242, opened 2022-06-21
Start `TextClassificationSingle` task

Chandu-4444:
A PR to check up on my `TextClassificationTask` implementation. Currently it only contains OneHot, but I'll eventually add up the text-specific `Encoding`s.


Chandu-4444:
Updated in #245 .

PR #245, opened 2022-06-25
Text classification task

Chandu-4444:
Implemented the text encodings and `TextClassificationTask`. Still need to do numericalization and a few more encodings. Then will focus on writing docstrings and tests.


Chandu-4444:
```julia

julia> using Revise

julia> using FastAI
[ Info: Precompiling FastAI [5d0beca9-ade8-49ae-ad0b-a3cf890e669f]

julia> data, blocks = load(datarecipes()["imdb"])
((mapobs(loadfile, ObsView(::MLDatasets.FileDataset{typeof(identity), String}, ::Vector{Int64})), mapobs(parentname, ObsView(::MLDatasets.FileDataset{typeof(identity), String}, ::Vector{Int64}))), (Paragraph(), WARNING: both TextAnalysis and FluxTraining export "fit!"; uses of it in module FastAI must be qualified
Label{String}(["neg", "pos"])))

julia> task = FastAI.TextClassficationSingle(blocks, data)
SupervisedTask(Paragraph -> Label{String})

julia> input, target = getobs(data, 2)
("Airport '77 starts as a brand new luxury 747 plane is loaded up with valuable paintings & such belonging to rich businessman Philip Stevens (James Stewart) who is flying them & a bunch of VIP's to his estate in preparation of it being opened to the public as a museum, also on board is Stevens daughter Julie (Kathleen Quinlan) & her son. The luxury jetliner takes off as planned but mid-air the plane is hi-jacked by the co-pilot Chambers (Robert Foxworth) & his two accomplice's Banker (Monte Markham) & Wilson (Michael Pataki) who knock the passengers & crew out with sleeping gas, they plan to steal the valuable cargo & land on a disused plane strip on an isolated island but while making his descent Chambers almost hits an oil rig in the Ocean & loses control of the plane sending it crashing into the sea where it sinks to the bottom right bang in the middle of the Bermuda Triangle. With air in short supply, water leaking in & having flown over 200 miles off course the problems mount for the survivor's as they await help with time fast running out...<br /><br />Also known under the slightly different tile Airport 1977 this second sequel to the smash-hit disaster thriller Airport (1970) was directed by Jerry Jameson & while once again like it's predecessors I can't say Airport '77 is any sort of forgotten classic it is entertaining although not necessarily for the right reasons. Out of the three Airport films I have seen so far I actually liked this one the best, just. It has my favourite plot of the three with a nice mid-air hi-jacking & then the crashing (didn't he see the oil rig?) & sinking of the 747 (maybe the makers were trying to cross the original Airport with another popular disaster flick of the period The Poseidon Adventure (1972)) & submerged is where it stays until the end with a stark dilemma facing those trapped inside, either suffocate when the air runs out or drown as the 747 floods or if any of the doors are opened & it's a decent idea that could have made for a great little disaster flick but bad unsympathetic character's, dull dialogue, lethargic set-pieces & a real lack of danger or suspense or tension means this is a missed opportunity. While the rather sluggish plot keeps one entertained for 108 odd minutes not that much happens after the plane sinks & there's not as much urgency as I thought there should have been. Even when the Navy become involved things don't pick up that much with a few shots of huge ships & helicopters flying about but there's just something lacking here. George Kennedy as the jinxed airline worker Joe Patroni is back but only gets a couple of scenes & barely even says anything preferring to just look worried in the background.<br /><br />The home video & theatrical version of Airport '77 run 108 minutes while the US TV versions add an extra hour of footage including a new opening credits sequence, many more scenes with George Kennedy as Patroni, flashbacks to flesh out character's, longer rescue scenes & the discovery or another couple of dead bodies including the navigator. While I would like to see this extra footage I am not sure I could sit through a near three hour cut of Airport '77. As expected the film has dated badly with horrible fashions & interior design choices, I will say no more other than the toy plane model effects aren't great either. Along with the other two Airport sequels this takes pride of place in the Razzie Award's Hall of Shame although I can think of lots of worse films than this so I reckon that's a little harsh. The action scenes are a little dull unfortunately, the pace is slow & not much excitement or tension is generated which is a shame as I reckon this could have been a pretty good film if made properly.<br /><br />The production values are alright if nothing spectacular. The acting isn't great, two time Oscar winner Jack Lemmon has said since it was a mistake to star in this, one time Oscar winner James Stewart looks old & frail, also one time Oscar winner Lee Grant looks drunk while Sir Christopher Lee is given little to do & there are plenty of other familiar faces to look out for too.<br /><br />Airport '77 is the most disaster orientated of the three Airport films so far & I liked the ideas behind it even if they were a bit silly, the production & bland direction doesn't help though & a film about a sunken plane just shouldn't be this boring or lethargic. Followed by The Concorde ... Airport '79 (1979).", "neg")

julia> encodesample(task, Training(), (input, target))
(Any[25000, 633779, 93, 1220, 135, 39, 322, 85, 93, 69  …  339, 1812, 22, 633779, 373, 633779, 633779, 7, 633779, 93], Float32[1.0, 0.0])

```

src/FastAI.jl
```diff
@@ -24,10 +24,16 @@ using StaticArrays
 using Setfield
 using ShowCases
 using Tables
+using TextAnalysis
```
lorenzoh, line 27: Can you add the symbols used here as you did for WordTokenizers? Then it's easier to see where variables come from and the namespace isn't polluted

src/FastAI.jl
```diff
@@ -24,10 +24,16 @@ using StaticArrays
...
 import Test
 import UnicodePlots
 using Statistics
 using InlineTest
+using DataStructures
```
lorenzoh, line 32: Same as for TextAnalysis

src/Textual/Textual.jl
```diff
@@ -17,18 +17,43 @@ using ..FastAI.Datasets
...
     _registerrecipes()
+    foreach(values(_tasks)) do t
+        if !haskey(FastAI.learningtasks(), t.id)
+            push!(FastAI.learningtasks(), t)
+        end
+    end
 end
 
-export Paragraph
+export Paragraph, TokenVector, TextClassficationSingle, Sanitize, Tokenize, replace_all_caps, replace_sentence_case, convert_lowercase
```
lorenzoh, line 58: I think we shouldn't export the replace_* and convert_* functions for now as the high-level exported interface should be used via the encodings or tasks.

src/Textual/blocks/text.jl
```diff
@@ -27,3 +27,11 @@ struct Paragraph <: Block end
 
 FastAI.checkblock(::Paragraph, ::String) = true
 FastAI.mockblock(::Paragraph) = randstring(" ABCEEFGHIJKLMNOPQESRUVWXYZ 1234567890 abcdefghijklmnopqrstynwxyz\n\t.,", rand(10:40))
+
+struct TokenVector <: Block end
+
+FastAI.checkblock(::TokenVector, ::Vector{String}) = true
```
lorenzoh, line 33: Let's call this block `Tokens`, to make clear it's multiple tokens. These blocks representing intermediate steps in the data pipeline also won't be exported (same as e.g. `ImageTensor`), so a short name is fine I think.

src/Textual/encodings/textpreprocessing.jl
```diff
@@ -0,0 +1,124 @@
...
+Encodes
+- `Paragraph` -> `Paragraph`
+
+"""
+struct Sanitize <: Encoding
+    tfms
+end
+
+function Sanitize()
+    base_tfms = [
```
lorenzoh, line 16: let's make `base_tfms` a constant in the module scope, i.e. `const DEFAULT_SANITIZERS`.

Then this definition becomes `Sanitize() = Sanitize(DEFAULT_SANITIZERS)`

src/Textual/encodings/textpreprocessing.jl
```diff
@@ -0,0 +1,124 @@
...
+        remove_punctuations,
+        basic_preprocessing,
+        remove_extraspaces,
+    ]
+    return Sanitize(base_tfms)
+end
+
+function encodedblock(p::Sanitize, block::Paragraph)
+    return block
+end
```
lorenzoh, line 29:
```suggestion
encodedblock(::Sanitize, block::Paragraph) = block
```

src/Textual/encodings/textpreprocessing.jl
```diff
@@ -0,0 +1,124 @@
...
+    end
+    obs
+end
+
+struct Tokenize <: Encoding
+    tfms
+end
+
+function Tokenize()
+    base_tfms = [
```
lorenzoh, line 43: Do we ever need multiple tokenizers? I'm not super familiar with it.

If yes, it should be the same as above, i.e. make this a `DEFAULT_TOKENIZERS`

src/FastAI.jl
lorenzoh, line 32:
```suggestion
using DataStructures: OrderedDict
```

src/Textual/tasks/classification.jl
```diff
@@ -0,0 +1,56 @@
...
+    blocks=(Paragraph, Label),
+    category="supervised",
+    description="""
+      Single-label text classification task where every text has a single
+      class label associated with it.
+      """,
+    package=@__MODULE__,
+)
+
+# ## Tests
```
lorenzoh, line 34: Do these work?

lorenzoh (commented):
Sorry for taking so long, but I've now had the chance to leave some comments. The implementation of the encodings looks good, I've left some minor comments here and there.

Would be great to have some tests for encodings, using both `FastAI.testencoding` and some specific input-output pairs for all the encodings.

src/Textual/tasks/classification.jl
Chandu-4444: Yes, I checked it and it's working fine.

ToucheSir: Then I would leave them uncommented so CI can do its work :)

src/Textual/encodings/textpreprocessing.jl
```diff
@@ -0,0 +1,124 @@
...
+
+function encode(p::Tokenize, context, block::Paragraph, obs)
+    for tfm in values(p.tfms)
+        obs = tfm(obs)
+    end
+    obs
+end
+
+function computevocabulary(data)
+    lookup_table = Dict()
```
ToucheSir, line 61:
```suggestion
    lookup_table = Dict{String,Int{()
```

src/Textual/encodings/textpreprocessing.jl
```diff
@@ -0,0 +1,124 @@
...
+
+    enc1 = Sanitize()
+    sanitized_Data = mapobs((i) -> encode(enc1, Training(), Paragraph(), getobs(data, i)[1]), 1:25000)
```
ToucheSir, line 64:
```suggestion
    sanitized_Data = map(i -> encode(enc1, Training(), Paragraph(), getobs(data, i)[1]), 1:25000)
```
This upper bound is a little odd to me. Why is it fixed at 250k?

src/Textual/encodings/textpreprocessing.jl
```diff
@@ -0,0 +1,124 @@
...
+
+    enc2 = Tokenize()
+    tokenized_data = mapobs((i) -> encode(enc2, Training(), Paragraph(), getobs(sanitized_Data, i)), 1:25000)
```
ToucheSir, line 67: Ditto here.

src/Textual/encodings/textpreprocessing.jl
```diff
@@ -0,0 +1,124 @@
...
+
+function encode(p::EmbedVocabulary, context, block::TokenVector, obs)
+    vocabulary = p.vocab
+
+    return_vect = []
+
+    for token in obs
+        push!(return_vect, getindex(vocabulary, token))
+    end
+    return return_vect
```
ToucheSir, line 103:
```suggestion
    return [vocabulary[token] for token in obs]
```

src/Textual/encodings/textpreprocessing.jl
```diff
@@ -0,0 +1,124 @@
...
+
+Encodes `Paragraph`s by applying various textual transforms.
+
+
+Encodes
+- `Paragraph` -> `Paragraph`
+
+"""
+struct Sanitize <: Encoding
+    tfms
```
ToucheSir, line 12: Is `tfms` always a Vector? This can probably be constrained further.

src/Textual/tasks/classification.jl
Chandu-4444: Sure! now that I've changed the structure of `block`s a bit I need to modify these.

src/Textual/encodings/textpreprocessing.jl
Chandu-4444: That is the total number of samples in `data`. I should have used `numobs(data)` instead. I'll change this now.

src/Textual/encodings/textpreprocessing.jl
Chandu-4444: Hmm.. Not always a vector and it can be empty too. But if one transform is used that'll probably be used along with a couple of other transforms like remove_puncts, stripping case etc.

ToucheSir: What can it be other than a vector? Can you enumerate the possible types?

src/Textual/encodings/textpreprocessing.jl
ToucheSir: Can you `mapobs` over `data` directly then?

src/Textual/encodings/textpreprocessing.jl
Chandu-4444: I mean it can either be empty (empty vector?) or a vector. For example, in the case of text generation, we don't have any sanitization for the input data and here sanitization step can be skipped. But yes! it'll always be a vector then.

src/Textual/encodings/textpreprocessing.jl
Chandu-4444: I didn't get that. Can you please elaborate?

ToucheSir: Instead of mapping over the indices of data and calling `getobs` for each index, why not just call `mapobs` with data?

lorenzoh:
This is ready to merge, right? @darsnack

darsnack (approved):

PR #267, opened 2022-11-26
Add a feature registry for models

lorenzoh:
### PR Checklist

- [x] Tests are added
- [x] Documentation (this is an internal change for now, docs will be added in follow-up when functionality is made available and used in domain package)


lorenzoh:
@darsnack @theabhirath would love to get feedback on the API. Anything unclear or an important feature missing, let me know!

PR #282, opened 2023-07-10
Remove extra type parameter

cpfiffer:
This is a super simple PR to remove an extraneous type parameter that throws a warning on Julia 1.9>=.

Here's the relevant error:

```
┌ FastAI [5d0beca9-ade8-49ae-ad0b-a3cf890e669f]
│  WARNING: method definition for mockblock at /home/cameron/.julia/packages/FastAI/f27xT/src/encodings/onehot.jl:37 declares type variable N but does not use it.
└  
```


darsnack:
Thanks!

